PR #1, opened 2021-07-07
Move mypy config to pyproject.toml; Fix autoformatting in pre-commit

dynobo:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,6 +4,7 @@ repos:
     rev: 21.5b1
     hooks:
       - id: black
+        exclude: ^blackboxopt/examples/
       - id: black
         # Check shorter line-length for examples (better with mkdocs)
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -12,15 +13,17 @@ repos:
     rev: 5.8.0
     hooks:
       - id: isort
+        args: ["--profile", "black"]
+        exclude: ^blackboxopt/examples/
```
LGro, line 17: Can we then have the same hook again that covers the examples directory? Otherwise, for unsorted imports in the examples, I'd expect to run into issues only when the matching action is run.

LGro (approved):

.pre-commit-config.yaml
dynobo: We have? (L15-17) Or do you mean something differently?

PR #4, opened 2021-07-07
Add py.typed to enable type discovery by mypy etc.

dynobo:
This file indicates that we have [type annotated source code](https://mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages). It makes mypy discover the types correctly, in case blackboxopt get's installed as dependency)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,143 @@
+# Custom
+.vscode/
+poetry.toml
+dask-worker-space/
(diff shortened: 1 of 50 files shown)


LGro (approved):

dynobo (commented):
- [ ] Rebase before merge

PR #6, opened 2021-07-09
Add GitHub actions

dynobo:


LGro (approved):
:rocket:

dynobo:
For the tests to pass, we need [parameterspace PR#19](https://github.com/boschresearch/parameterspace/pull/19).

poetry.lock
```diff
@@ -0,0 +1,1631 @@
+[[package]]
```
dynobo, line 1: - [ ] Before merging, do `poetry update` (requires [parameterspace #19](https://github.com/boschresearch/parameterspace/pull/19))

LGro: I'm afraid it's not as simple as adding `__iter__`. What this linting issue highlights is that BOHB is not ok with any `SearchSpace` but actually requires a `ParameterSpace`.
I'd recommend to mark the current linting issue as skip and create an issue for us to figure out separately from this PR how to deal with optimizers that requires specific knowledge about the parameter types. Maybe we need to add them to the base `SearchSpace` implementation in some way as well or allow individual optimizers to be more specific about the search space definition they allow.

dynobo: Created issue #10 for this.
I will set the line to be ignored by pylint. We should re-enable that, after the issue is fixed.

dynobo:
The solution will be discussed and resolved as part of `blackboxopt` [issue 10](https://github.com/boschresearch/blackboxopt/issues/10)

PR #13, opened 2021-07-15
Publish docs to gh pages

dynobo:
See result at [boschresearch.github.io/blackboxopt](https://boschresearch.github.io/blackboxopt/).

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -19,8 +19,10 @@ jobs:
       - name: Sorted Imports
         if: always()
         run: |
-          poetry run isort --check-only --profile black --line-length 88 --skip-glob blackboxopt/examples/* .
-          poetry run isort --check-only --profile black --line-length 80 blackboxopt/examples/
+          poetry run isort --check-only --profile black --line-length 88 --skip-glob \
+            blackboxopt/examples/* .
+          poetry run isort --check-only --profile black --line-length 80 \
+            blackboxopt/examples/
       - name: Type Checks
         if: always()
         run: poetry run mypy --config-file pyproject.toml .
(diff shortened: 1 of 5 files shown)


LGro:
How do you feel about going with `mkdocs-versioning` and doing `mkdocs-versioning deploy` to get https://boschresearch.github.io/blackboxopt/1.0.1 like URLs and retain documentation for older versions?

LGro (approved):

dynobo:
> How do you feel about going with `mkdocs-versioning` and doing `mkdocs-versioning deploy` to get https://boschresearch.github.io/blackboxopt/1.0.1 like URLs and retain documentation for older versions?

Versioning is a good idea, but instead `mkdocs-versioning` let's go with `mike` which [seems to be the standard for mkdocs-material](https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/#versioning)

.github/workflows/python.yaml
```diff
@@ -39,12 +39,34 @@ jobs:
...
-      - name: Release
+      - name: Build & release
         env:
           PYPI_TOKEN: ${{ secrets.PYPI_TOKEN }}
         run: |
           poetry config pypi-token.pypi $PYPI_TOKEN
           poetry publish --build
+  docs:
+    needs: test
+    # if: github.ref == 'refs/heads/main'
```
LGro, line 53: Do I understand correctly that for the desired state we would only run this on tagged releases?

LGro (approved):

LGro:
Reminder for us to remove the commented out release condition after we have confirmed the docs deployment works.

.github/workflows/python.yaml
dynobo: As discussed, I changed to same condition as we have for pypi deployment: if tag starts with `v`.

PR #16, opened 2021-08-06
Add pydocstyle linter

dynobo:
Adds [pydocstyle](https://github.com/PyCQA/pydocstyle) for linting google-styled docs. Check is added to gh action, but with the "continue-on-error" switch, as we have to fix quite a bunch of issues first (and/or adjust the linter config).

Hints:
- Contributors using `vscode` can easily enable issue-highlighting by enabling `python.linting.pydocstyleEnabled` in the settings UI.
- In CLI just run `pydocstyle` (or `pydocstyle --explain` for more details)


sfalkner (approved):
Thanks @dynobo!

PR #17, opened 2021-08-31
Generic raise on unknown or incomplete

dynobo:


LGro (approved):
Looks good, thanks @dynobo!

PR #19, opened 2021-09-03
Switch to report evaluations in batches

dynobo:
This PR switches from reporting evaluations individually to allow reporting batches (lists) of evaluations, as discussed with @LGro in person. 

- Allows better performance in cases where we can fit a model with all evaluations in one run instead fitting it multiple times for every individual evaluation. @sfalkner : Would this be the case for e.g. BOHB? If yes, feel free to contribute an PR. :slightly_smiling_face:
- The `StagedIterationOptimizer` and the `sequential` optimization loop are not (yet?) adopted to leverage this functionality and just keep processing the evaluations individually.
- Individual evaluations still can be reported as a list of one item. (I propose to not allow `Union[List[Evaluation], Evaluation]` to keep the code and tests etc. more simple. What do you think?)

### DIFF
--- a/blackboxopt/base.py
+++ b/blackboxopt/base.py
@@ -104,14 +104,14 @@ class Optimizer(abc.ABC):
         """
 
     @abc.abstractmethod
-    def report_evaluation(self, evaluation: Evaluation) -> None:
-        """Report an evaluated evaluation specification.
+    def report_evaluations(self, evaluations: Iterable[Evaluation]) -> None:
+        """Report multiple evaluated evaluation specifications.
 
         NOTE: Not all optimizers support reporting results for evaluation specifications
         that were not proposed by the optimizer.
 
         Args:
-            evaluation: An evaluated evaluation specification.
+            evaluations: An iterable of evaluated evaluation specifications.
         """
 
 
@@ -133,14 +133,14 @@ class SingleObjectiveOptimizer(Optimizer):
         super().__init__(search_space=search_space, seed=seed)
         self.objective = objective
 
-    def report_evaluation(self, evaluation: Evaluation) -> None:
-        raise_on_unknown_or_incomplete(
-            exception=ObjectivesError,
-            known=[self.objective.name],
-            reported=evaluation.objectives.keys(),
-        )
-
-        super().report_evaluation(evaluation)
+    def report_evaluations(self, evaluations: Iterable[Evaluation]) -> None:
+        for evaluation in evaluations:
+            raise_on_unknown_or_incomplete(
+                exception=ObjectivesError,
+                known=[self.objective.name],
+                reported=evaluation.objectives.keys(),
+            )
+        super().report_evaluations(evaluations)
 
 
 class MultiObjectiveOptimizer(Optimizer):
@@ -163,11 +163,12 @@ class MultiObjectiveOptimizer(Optimizer):
         _raise_on_duplicate_objective_names(objectives)
         self.objectives = objectives
 
-    def report_evaluation(self, evaluation: Evaluation) -> None:
-        raise_on_unknown_or_incomplete(
-            exception=ObjectivesError,
-            known=[o.name for o in self.objectives],
-            reported=evaluation.objectives.keys(),
-        )
+    def report_evaluations(self, evaluations: Iterable[Evaluation]) -> None:
+        for evaluation in evaluations:
+            raise_on_unknown_or_incomplete(
+                exception=ObjectivesError,
+                known=[o.name for o in self.objectives],
+                reported=evaluation.objectives.keys(),
+            )
 
-        super().report_evaluation(evaluation)
+        super().report_evaluations(evaluations)
--- a/blackboxopt/base.py
+++ b/blackboxopt/base.py
@@ -105,9 +105,7 @@ class Optimizer(abc.ABC):
         """
 
     @abc.abstractmethod
-    def report_evaluations(
-        self, evaluations: Union[Evaluation, Iterable[Evaluation]]
-    ) -> None:
+    def report(self, evaluations: Union[Evaluation, Iterable[Evaluation]]) -> None:
         """Report one or more evaluated evaluation specifications.
 
         NOTE: Not all optimizers support reporting results for evaluation specifications
@@ -137,9 +135,7 @@ class SingleObjectiveOptimizer(Optimizer):
         super().__init__(search_space=search_space, seed=seed)
         self.objective = objective
 
-    def report_evaluations(
-        self, evaluations: Union[Evaluation, Iterable[Evaluation]]
-    ) -> None:
+    def report(self, evaluations: Union[Evaluation, Iterable[Evaluation]]) -> None:
         if isinstance(evaluations, Evaluation):
             evaluations = [evaluations]
 
@@ -150,7 +146,7 @@ class SingleObjectiveOptimizer(Optimizer):
                 reported=evaluation.objectives.keys(),
             )
 
-        super().report_evaluations(evaluations)
+        super().report(evaluations)
 
 
 class MultiObjectiveOptimizer(Optimizer):
@@ -173,9 +169,7 @@ class MultiObjectiveOptimizer(Optimizer):
         _raise_on_duplicate_objective_names(objectives)
         self.objectives = objectives
 
-    def report_evaluations(
-        self, evaluations: Union[Evaluation, Iterable[Evaluation]]
-    ) -> None:
+    def report(self, evaluations: Union[Evaluation, Iterable[Evaluation]]) -> None:
         if isinstance(evaluations, Evaluation):
             evaluations = [evaluations]
 
@@ -186,4 +180,4 @@ class MultiObjectiveOptimizer(Optimizer):
                 reported=evaluation.objectives.keys(),
             )
 
-        super().report_evaluations(evaluations)
+        super().report(evaluations)
--- a/blackboxopt/base.py
+++ b/blackboxopt/base.py
@@ -5,7 +5,6 @@
 
 import abc
 import collections
-from collections.abc import Iterable as abc_Iterable
 from dataclasses import dataclass
 from typing import Iterable, List, Type, Union
 
@@ -36,6 +35,17 @@ class ConstraintsError(ValueError):
 
 class ContextError(ValueError):
     """Raised on incomplete or missing context information."""
+
+
+class EvaluationsError(ValueError):
+    """Raised on invalid evaluations.
+
+    The problematic evaluations are passed in the `evaluations` attribute.
+    """
+
+    def __init__(self, message: str, evaluations: List[Evaluation]):
+        self.message = message
+        self.evaluations = evaluations
 
 
 @dataclass
--- a/blackboxopt/base.py
+++ b/blackboxopt/base.py
@@ -5,8 +5,9 @@
 
 import abc
 import collections
+from collections.abc import Iterable as abc_Iterable
 from dataclasses import dataclass
-from typing import Dict, Iterable, List, Optional, Type
+from typing import Iterable, List, Type, Union
 
 from parameterspace.base import SearchSpace
 
@@ -104,14 +105,17 @@ class Optimizer(abc.ABC):
         """
 
     @abc.abstractmethod
-    def report_evaluations(self, evaluations: Iterable[Evaluation]) -> None:
-        """Report multiple evaluated evaluation specifications.
+    def report_evaluations(
+        self, evaluations: Union[Evaluation, Iterable[Evaluation]]
+    ) -> None:
+        """Report one or more evaluated evaluation specifications.
 
         NOTE: Not all optimizers support reporting results for evaluation specifications
         that were not proposed by the optimizer.
 
         Args:
-            evaluations: An iterable of evaluated evaluation specifications.
+            evaluations: A single evaluated evaluation specifications, or an iterable
+            of many.
         """
 
 
@@ -133,13 +137,19 @@ class SingleObjectiveOptimizer(Optimizer):
         super().__init__(search_space=search_space, seed=seed)
         self.objective = objective
 
-    def report_evaluations(self, evaluations: Iterable[Evaluation]) -> None:
+    def report_evaluations(
+        self, evaluations: Union[Evaluation, Iterable[Evaluation]]
+    ) -> None:
+        if isinstance(evaluations, Evaluation):
+            evaluations = [evaluations]
+
         for evaluation in evaluations:
             raise_on_unknown_or_incomplete(
                 exception=ObjectivesError,
                 known=[self.objective.name],
                 reported=evaluation.objectives.keys(),
             )
+
         super().report_evaluations(evaluations)
 
 
@@ -163,7 +173,12 @@ class MultiObjectiveOptimizer(Optimizer):
         _raise_on_duplicate_objective_names(objectives)
         self.objectives = objectives
 
-    def report_evaluations(self, evaluations: Iterable[Evaluation]) -> None:
+    def report_evaluations(
+        self, evaluations: Union[Evaluation, Iterable[Evaluation]]
+    ) -> None:
+        if isinstance(evaluations, Evaluation):
+            evaluations = [evaluations]
+
         for evaluation in evaluations:
             raise_on_unknown_or_incomplete(
                 exception=ObjectivesError,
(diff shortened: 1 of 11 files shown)


blackboxopt/optimizers/staged/optimizer.py
```diff
@@ -46,20 +46,27 @@ def __init__(
...
+        missing_ids_count = sum(e.optimizer_info.get("id") is None for e in evaluations)
+        if missing_ids_count > 0:
             raise ValueError(
-                "Missing evaluation specification ID in optimizer info. Did you try to "
-                + "report an evaluation for a configuration which the optimizer did not"
-                " pick? This is not supported at the moment."
+                f"{missing_ids_count} evaluation(s) miss an evaluation specification ID"
+                + "in the optimizer info. "
+                + "Did you try to report evaluations for configurations which the "
+                + "optimizer did no pick? This is not supported at the moment."
```
dynobo, line 58: I'm not sure, what the best behavior here. I considered three options:
- process all evaluations that are valid, skip only the invalid ones (ignore).
- process all evaluations until an invalid one, and ignore the following ones (abort).
- process none evaluations, even if only a single one is invalid (strict).

Here I proposed the "strict" version, because I think it gives the most control to user. Also it might be more difficult to _identify_ the invalid evaluations if some are processed and others are not. 

What do you think, @sfalkner, @LGro ?

LGro:
>     * Individual evaluations still can be reported as a list of one item. (I propose to not allow `Union[List[Evaluation], Evaluation]` to keep the code and tests etc. more simple. What do you think?)

The code and tests remain more simple that way, which I appreciate from a developer and maintainer perspective. From a user perspective it seems like the UX of having to wrap individual evaluations as `[evaluation]` questionable, though. Do we know how other frameworks handle this "one or multiple" paradigm?

sfalkner:
I would second Lukas' argument and would also think that the `Union` solution would actually be better from the user perspective.

dynobo:
I did some research, but the opinions on this topic vary a lot. I know that pandas uses the `Union` style to allow lists as well as single items, and I like that, too.

I'd prefer a single function which accepts a `Union` over `report_evaluation` + `report_evaluations`, because they look so similar and `report_single_evaluation` would be cumbersome. I'll update the PR.

But  could you two please also comment on the (more critical) [question regarding the behavior](https://github.com/boschresearch/blackboxopt/pull/19#discussion_r701964129)?

sfalkner:
I personally would prefer the `ignore` one, but would emit a warning for every evaluation that couldn't be processed or a summary warning that X evaluations have been dropped. That way, it should be easy to simply put in results from other runs without filtering out the incomplete evaluations that don't have a loss. The warning(s) still make the behavior transparent to the user.

blackboxopt/base.py
```diff
@@ -133,14 +137,20 @@ def __init__(
...
+
+        for evaluation in evaluations:
+            raise_on_unknown_or_incomplete(
+                exception=ObjectivesError,
+                known=[self.objective.name],
+                reported=evaluation.objectives.keys(),
+            )
 
-        super().report_evaluation(evaluation)
+        super().report_evaluations(evaluations)
```
dynobo, line 153: By the way: Am I correct that this super call doesn't make any sense? As the `super()` class is in this case an not implemented abstract class anyway?

LGro:
I currently see the following two main base scenarios.
1. I expect all my evaluations to be accepted.
2. I expect that not all my evaluations are going to be accepted.

For (1) my clear preference is a strict handling of unaccepted evaluations i.e. raising an exception that contains information about the rejected evaluation(s).

For (2) can see the benefit of us issuing warnings so the user can react, but it also seems too inflexible and unnecessarily difficult to handle programmatically on the user side. Instead, I'd prefer to be strict as well, i.e. raising an exception that contains information about the rejected evaluation(s).

This leaves me in charge to set my expectations. Either I surround the `report_evaluation` with exception handling because I expect this to happen or I don't and then notice because it fails. Having warnings and not failing seems dangerous because it implies expectations that might not be consistent across our user base. I need to look for warnings to notice, so it might well happen that some evaluations are dropped without the user noticing when they actually wouldn't expect it to happen.

A recommended setup like that would also allow to easily handle rejected evaluations on the user side, e.g.:
```python
try:
    report_evaluations(my_evals)
except EvaluationsError as e:
    # I do something meaningful with the rejected evaluations
    custom_book_keeping(e.evaluations)
```

LGro:
I agree with @sfalkner in the sense that I wouldn't reject all evaluations when only some of them are incompatible.

dynobo:
> [for the single method, we could also just make that `optimizer.report(evaluations)`](https://github.com/boschresearch/blackboxopt/pull/20#issuecomment-914157172)

Yep, just `report()` is a great idea! I also prefer it over having two methods with longer names. 
I wouldn't introduce the `strict_reject_all` switch before it's actually required.

LGro:
Moving the misplaced comment that @dynobo answered to here for reference:

One thing that comes to mind when I think about the indeed rather confusing similarity of report_evaluation and report_evaluations, we could also make them report_single_evaluation and report report_multiple_evaluations.

For the single, combined method approach, we could also just make that optimizer.report(evaluations); the current redundancy of method name and argument seems unnecessary, since there are usually not different things to report.

blackboxopt/base.py
dynobo: Or what this done on purpose, @LGro ?

blackboxopt/base.py
```diff
@@ -37,6 +38,17 @@ class ContextError(ValueError):
     """Raised on incomplete or missing context information."""
 
 
+class EvaluationsError(ValueError):
+    """Raised on invalid evaluations (e.g. missing evaluation specificiation id).
```
LGro, line 42: What is an evaluation specification ID?

blackboxopt/base.py
```diff
@@ -104,14 +116,15 @@ def get_evaluation_specification(self) -> EvaluationSpecification:
         """
 
     @abc.abstractmethod
-    def report_evaluation(self, evaluation: Evaluation) -> None:
-        """Report an evaluated evaluation specification.
+    def report(self, evaluations: Union[Evaluation, Iterable[Evaluation]]) -> None:
```
LGro, line 119: You mentioned having seen that paradigm with pandas. Do they use singular or plural names for the combined arguments that allow both single and multiples?

blackboxopt/optimizers/staged/optimizer.py
```diff
@@ -46,20 +46,33 @@ def __init__(
...
 
-        self.pending_configurations.pop(str(evaluation_specification_id))
-        idx = self.evaluation_uuid_to_iteration.pop(str(evaluation_specification_id))
-        self.iterations[idx].digest_evaluation(evaluation_specification_id, evaluation)
+        if invalid_evaluations:
+            raise EvaluationsError(
+                f"{len(invalid_evaluations)} evaluation(s) got rejected because of a "
+                "missing an evaluation specification ID in the optimizer info.",
+                evaluations=invalid_evaluations,
+            )
```
LGro, line 75: To combat that separation of where the error is detected and where it's reported, let's rename the variable to `evaluations_without_ids`, because otherwise one might happily add evaluations that are invalid for other reasons to the variable and get a mismatched error message.

Unrelated, the error message should read "[...] because of missing an evaluation [...]"

blackboxopt/optimizers/testing.py
```diff
@@ -124,7 +122,8 @@ def is_deterministic_with_fixed_seed(optimizer_class, optimizer_kwargs: dict) ->
         )
 
         es1 = opt.get_evaluation_specification()
-        opt.report_evaluation(es1.create_evaluation(objectives={"loss": 0.42}))
+        evaluation1 = es1.create_evaluation(objectives={"loss": 0.42})
+        opt.report([evaluation1])
```
LGro, line 126: Is this intended to cover testing the batch report functionality?
I'm wondering whether it might make sense to add a separate test for that. Because if the reproducibility test fails due to an optimizer not supporting report many evaluations, this is unnecessarily hard for the developer to find out, when instead there could be a `handles_reporting_evaluations_list` which immediately from its name suggests what might have gone wrong when it fails.

blackboxopt/optimization_loops/dask_distributed.py
```diff
@@ -163,12 +163,12 @@ def run_optimization_loop(
...
         for evaluation in dask_scheduler.check_for_results(timeout_s=20):
-            optimizer.report_evaluation(evaluation)
+            optimizer.report(evaluation)
             evaluations.append(evaluation)
 
     while dask_scheduler.has_running_jobs():
         for evaluation in dask_scheduler.check_for_results(timeout_s=20):
-            optimizer.report_evaluation(evaluation)
+            optimizer.report(evaluation)
             evaluations.append(evaluation)
```
LGro, line 172: Could this become the following? And if so, what do you prefer?
```python
    while dask_scheduler.has_running_jobs():
        new_evaluations = dask_scheduler.check_for_results(timeout_s=20)
        optimizer.report(new_evaluations)
        evaluations.extend(new_evaluations)
```

blackboxopt/base.py
```diff
@@ -163,11 +180,15 @@ def __init__(
...
+        if isinstance(evaluations, Evaluation):
+            evaluations = [evaluations]
+
+        for evaluation in evaluations:
+            raise_on_unknown_or_incomplete(
+                exception=ObjectivesError,
+                known=[o.name for o in self.objectives],
+                reported=evaluation.objectives.keys(),
+            )
```
LGro, line 192: It seems to me we are being inconsistent with the different failure modes for evaluations. In case of objectives errors we don't accept any evaluations if one is off. In case of the missing IDs in BOHB, we accept all evaluations with IDs and only reject / raise on the ones that have issues.

blackboxopt/base.py
```diff
@@ -37,6 +38,17 @@ class ContextError(ValueError):
...
 
+class EvaluationsError(ValueError):
+    """Raised on invalid evaluations (e.g. missing evaluation specificiation id).
+
+    The problematic evaluations are passed in the `evaluations` attribute.
+    """
+
+    def __init__(self, message, evaluations):
+        self.message = message
+        self.evaluations = evaluations
```
LGro, line 49: Let's add type hints here
```python
    def __init__(self, message: str, evaluations: List[Evaluation]):
        self.message = message
        self.evaluations = evaluations
```

I'm also wondering if, for more complex error handling, it might pay to return more specific error details with each evaluation (think something is wrong with the objectives in one evaluation and another one is missing an optimizer specific ID).

blackboxopt/base.py
dynobo: Plurals, I would do it the same (It's more intuitions that passing one evaluation is okay if many are also possible, than the other way round.)

blackboxopt/base.py
dynobo: True, what do you propose?

blackboxopt/optimizers/testing.py
dynobo: Actually, no, it's not intended to cover batch functionality, I just forgot to remove the list. ;-)
But adding a test for batch might be a good idea, so I'll do it :-)

blackboxopt/base.py
dynobo: Type hints: :+1: 
More complex information: I wouldn't do that in this PR. If you see a requirement for that, we should think about a solution which we can also apply to our other custom exceptions. What do you think?

blackboxopt/optimization_loops/dask_distributed.py
dynobo: Good catch! I think it should work and is like it more!

dynobo:
> ... However, we could still have a switch `strict_reject_all=False` for `report_evaluations` with that liberal default.
@LGro: After implementing your proposed switch to throw an exception, would you still like to see the `strict_reject_all` flag, or should we leave the appropriate handling to the user?

blackboxopt/optimizers/testing.py
```diff
@@ -133,6 +132,39 @@ def is_deterministic_with_fixed_seed(optimizer_class, optimizer_kwargs: dict) ->
...
+        objectives=[Objective("loss", False)],
+    )
+    evaluations = []
+    for _ in range(3):
+        es = opt.get_evaluation_specification()
+        evaluation = es.create_evaluation(objectives={"loss": 0.42})
+        evaluations.append(evaluation)
+
+    opt.report(evaluations)
+    return True
```
dynobo, line 165: @LGro: Do you think there is something more we can test, except that `report()` doesn't raise?

blackboxopt/base.py
LGro: I'd propose to follow what we decided in the previous discussion, to accept all evaluations that don't fail and then raise on the failed ones. Which links to [this comment](https://github.com/boschresearch/blackboxopt/pull/19/files#r704079665).

blackboxopt/optimizers/testing.py
LGro: This is sufficient :relaxed:

blackboxopt/optimization_loops/dask_distributed.py
```diff
@@ -162,13 +162,13 @@ def run_optimization_loop(
...
+        optimizer.report(new_evaluation)
+        evaluations.extend(new_evaluation)
 
     while dask_scheduler.has_running_jobs():
-        for evaluation in dask_scheduler.check_for_results(timeout_s=20):
-            optimizer.report_evaluation(evaluation)
-            evaluations.append(evaluation)
+        new_evaluation = dask_scheduler.check_for_results(timeout_s=20)
+        optimizer.report(new_evaluation)
+        evaluations.extend(new_evaluation)
```
LGro, line 172: Seems like this should be plural `new_evaluations` since it can be multiple and is always an iterable IIRC.

blackboxopt/base.py
LGro: See [here](https://github.com/boschresearch/blackboxopt/pull/19/files#r704078451) for the requirement.

blackboxopt/base.py
```diff
@@ -37,6 +37,20 @@ class ContextError(ValueError):
     """Raised on incomplete or missing context information."""
 
 
+class EvaluationsError(ValueError):
+    """Raised on invalid evaluations.
+
+    The problematic evaluations are passed in the `evaluations` attribute.
```
LGro, line 43: This needs to say "evaluations_with_errors".

blackboxopt/base.py
```diff
@@ -133,14 +148,23 @@ def __init__(
         super().__init__(search_space=search_space, seed=seed)
         self.objective = objective
 
-    def report_evaluation(self, evaluation: Evaluation) -> None:
-        raise_on_unknown_or_incomplete(
-            exception=ObjectivesError,
-            known=[self.objective.name],
-            reported=evaluation.objectives.keys(),
-        )
+    def report(self, evaluations: Union[Evaluation, Iterable[Evaluation]]) -> None:
```
LGro, line 151: Let's introduce a docstring to explain the behaviour when erroneous evaluations are reported, which ones are accepted and how the user gets to learn about the failed ones and what went wrong for them (incl. a "raises" docstring section).

tests/optimizers/bohb_test.py
```diff
@@ -82,6 +87,36 @@ def test_bohb_parallel():
...
+    # and triggers a warning on reporting:
+    invalid_evaluation = Evaluation(objectives={"loss": 42}, configuration={})
+    evaluations.append(invalid_evaluation)
+
+    assert len(opt.pending_configurations) == 3
+    with pytest.raises(EvaluationsError) as excinfo:
+        opt.report(evaluations)
+    assert len(opt.pending_configurations) == 0
+    assert excinfo.value.message.startswith("An error with one or more evaluation")
+    assert excinfo.value.evaluations_with_errors[0][0] == invalid_evaluation
```
LGro, line 117: How is this different from the matching reference test that BOHB is also subjected to?

tests/optimizers/bohb_test.py
```diff
@@ -82,6 +87,36 @@ def test_bohb_parallel():
...
+    )
+
+    evaluations = []
+    for i in range(3):
+        es = opt.get_evaluation_specification()
+        evaluation = es.create_evaluation(objectives={"loss": i})
+        evaluations.append(evaluation)
+
+    # Add an evaluation _not_ created by the optimizer to see if it get's skipped
+    # and triggers a warning on reporting:
```
LGro, line 108: Let's be more specific here and say that it's an evaluation that is missing specific details in the optimizer info, because we might well add support for reporting evaluations for custom specifications independent from this test.

LGro (approved):
Almost there (see two open comments), thanks @dynobo for your stamina :relaxed:

PR #20, opened 2021-09-06
remove context from specification

LGro:
Since we discussed to allow context to individually fix parts of the search space on demand when requesting new evaluation specifications from the optimizer, I'd propose to remove the `context` attribute from the specification and rather leave it to the user to know what they fixed and the optimizer to represent that information as part of the `optimizer_info` attribute. What do you think?

### DIFF
--- a/blackboxopt/evaluation.py
+++ b/blackboxopt/evaluation.py
@@ -35,14 +35,6 @@ class EvaluationSpecification(Mapping[str, Any]):
         metadata={"Description": "Creation time of the evaluation specificiation."},
     )
 
-    context: Optional[Dict[str, Any]] = field(
-        default=None,
-        metadata={
-            "Description": "Contextual information is what you can determine but not "
-            + "influence, like the environmental temperature."
-        },
-    )
-
     def keys(self):
         return self.__dataclass_fields__.keys()  # pylint: disable=no-member
 
--- a/tests/evaluation_test.py
+++ b/tests/evaluation_test.py
@@ -88,7 +88,7 @@ def test_to_json():
     spec_json = spec.to_json()
     assert spec_json == (
         '{"configuration": {"p1": 1.2}, "settings": {"fidelity": 1.0}, '
-        '"optimizer_info": {"id": 123}, "created_unixtime": 1.0, "context": null}'
+        '"optimizer_info": {"id": 123}, "created_unixtime": 1.0}'
     )
 
     result = Evaluation(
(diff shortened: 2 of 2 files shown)


dynobo:
I'm not 100% sure... The specification still is created for a certain context, isn't it? Then it seems to make sense to keep the context close to spec itself... And how would a user access the context information (for which the specification was created) inside the evaluation function?

sfalkner:
Thanks for bringing this up again. Our use of context here is not very consistent and maybe we could clarify a few things first, before we reach a conclusion.

For me, the context should be very clearly separated from the search space, because one is not effectively searching over it, but has given values for each evaluation. Fixing parts of the space for each requested specification would allow context (and actually quite a bit more, because one can choose different subsets of variables for every call). I am not sure if the additional freedom adds any value, but I think it could still be relevant to log which variables have been fixed for each evaluation. In the current form this would be done by the entry that you propose to remove, right?

A related question is whether the function we optimize actually takes the context as an input. The doc string of context mentions the ambient temperature as an example, which doesn't sound like an input to any process. I guess it comes down to how explicitly we want to be when it comes to separating the actual parameters from additional information.

In any case, in order to fully realize context in our code, the `get_evaluation_specification` method of the optimizer should take an extra argument for the context values, which is currently not the case. I think it would be helpful to first think about a solution for the context (if we actually want to use that any time soon) and how that fits into the API. This would hopefully give us a better idea how to translate that into code.

LGro:
I'd at least make it convention to have
```
evaluation.optimizer_info["context"] : List[str]
```
the question is, if we also want to make this top level information. And while we are gaining experience with context enabled optimizers, I'd propose to have it in the `optimizer_info` until we see that we'd strongly benefit from making this a top level attribute.

dynobo:
> And while we are gaining experience with context enabled optimizers, I'd propose to have it in the `optimizer_info` until we see that we'd strongly benefit from making this a top level attribute.

That totally makes sense to me. :-) Let's start more simple/flexible with the `optimizer_info` and let's see where `context` use cases lead us... :-)

dynobo (approved):

sfalkner:
But doesn't that still leave the question open how the context is relayed to the optimizer for the creation of the specification and whether or not the context is a subset of the search space?

LGro:
We indeed still need to agree on
> how the context is relayed to the optimizer for the creation of the specification
> whether or not the context is a subset of the search space

One setup that would allow us to do what we currently need is the following, albeit not being explicitly "context":  
Allow an optional argument `Dict[str, Any]` in `get_evaluation_specification` to fix any parameter by name to a specific value. We can define that argument in the optimizer base today or start with only defining it in the optimizers that support it, both are viable from an OOP viewpoint.

If we feel like we want to figure out something that is more explicitly contextual optimization, then this could mean we pass in a search and a context space during optimizer initialization.

LGro:
One thing that comes to mind when I think about the indeed rather confusing similarity of `report_evaluation` and `report_evaluations`, we could also make them `report_single_evaluation` and report `report_multiple_evaluations`.

dynobo:
> One thing that comes to [...]

Misplaced comment, correct? Replied [here](https://github.com/boschresearch/blackboxopt/pull/19#issuecomment-914164098) in #19 .

LGro:
As discussed in person, we keep the explicit context for transparency and still implement parameter fixing as a separate feature.

PR #23, opened 2021-09-13
replace time with datetime for unix timestamps

LGro:
Creates compatibility with `datetime` powered timestamp conversion, which creates six instead of seven decimal places for unix timestamps compared to `time.time()` and would thus cause equality checks to fail after conversion.


dynobo (approved):
Thanks, @LGro !

PR #26, opened 2021-10-05
Hotfix/bohb maximize objective

LGro:
Noticing that we were missing support for maximizing objectives in our staged iteration optimizers, I wanted to add a test to check that we can both minimize and maximize.
The current proposal is after optimizing across three fidelities 1, 3, 9 to compare the average loss or score of the highest fidelity model picks vs. the middle fidelity model picks and make sure their performance decreased or increased accordingly. This seems to do the job in most cases but doesn't seem extremely elegant. @sfalkner, I'd kindly ask for your input on this one, especially wrt a more elegant test condition.


tests/optimizers/bohb_test.py
```diff
@@ -250,3 +249,81 @@ def test_bohb_with_none_min_samples_in_model():
...
+
+def test_minimize_quadratic():
+    paramspace = ps.ParameterSpace()
+    paramspace.add(ps.ContinuousParameter("p1", (-2.0, 2.0)))
+    opt = BOHB(
+        paramspace,
+        Objective("loss", False),
+        min_fidelity=1.0,
+        max_fidelity=9.0,
+        num_iterations=10,
```
sfalkner, line 262: Here is my proposal to make  this test reliable (without seeding).
I would restrict BOHB to one iteration here.

tests/optimizers/bohb_test.py
```diff
@@ -250,3 +249,81 @@ def test_bohb_with_none_min_samples_in_model():
...
+        seed=42,
+    )
+
+    evals = []
+    for _ in range(60):
```
sfalkner, line 267: Then looping for `(9+3+1) = 10` configs to evaluate the whole iteration.

tests/optimizers/bohb_test.py
```diff
@@ -250,3 +249,81 @@ def test_bohb_with_none_min_samples_in_model():
...
+        es = opt.get_evaluation_specification()
+        e = es.create_evaluation(objectives={"loss": es.configuration["p1"] ** 2})
+        evals.append(e)
+        opt.report(e)
+
+    best_eval = min(evals, key=lambda e: e.objectives["loss"])
+    assert abs(best_eval.objectives["loss"]) < 0.1
```
sfalkner, line 274: Not sure if that is an ideal `assert`, as this regret bound might not be achieved in the first 9 trails.

tests/optimizers/bohb_test.py
```diff
@@ -250,3 +249,81 @@ def test_bohb_with_none_min_samples_in_model():
...
+
+    max_fidelity_loss_mean = np.mean(
```
sfalkner, line 276: If only one iteration has been completed, the `mean` could be a `min` and that would actually check that the right configs got promoted all the way to the highest fidelity.

tests/optimizers/bohb_test.py
```diff
@@ -250,3 +249,81 @@ def test_bohb_with_none_min_samples_in_model():
...
+        [
+            e.objectives["loss"]
+            for e in evals
+            if e.settings["fidelity"] == 3.0 and e.optimizer_info["model_based_pick"]
+        ]
+    )
+    assert max_fidelity_loss_mean < mid_fidelity_loss_mean
+
+
+def test_maximize_quadratic():
```
sfalkner, line 293: Same adjustments here, but with `min` -> `max`, I guess.

sfalkner (approved):

LGro:
Thanks a lot for the suggestion, @sfalkner. Your proposal is well suited to test the evaluation digestion in the iteration, which I missed entirely before. However, to also still cover the BOHB sampler, I added another test which is now more explicitly about the way maximization objective values are converted to losses.
Please give the two most recent commits another look and let me know via a :+1: on this comment when everything is alright.

PR #29, opened 2021-11-04
rename get_evaluation_specification to create_evaluation_specification

LGro:
For UX reasons, it makes sense to change away from `get_`, but instead of the now introduced `create_` we could also go for `propose_` if you prefer that, @sfalkner / @dynobo.


dynobo:
`propose_` :+1:

dynobo (approved):

sfalkner:
How about `generate_`?

PR #30, opened 2021-11-04
Inject additional js in html output of to_html()

dynobo:
Adds functionality to the html version of the visualization:
- Click on data point to toggle persistence of hover-box
- Click on text-line in hover-box to copy the value of the line into clipboard.

### DIFF
--- a/blackboxopt/visualizations/utils.py
+++ b/blackboxopt/visualizations/utils.py
@@ -179,6 +179,9 @@ def mask_pareto_efficient(costs: np.ndarray):
 def patch_plotly_io_to_html(method):
     """Patch `plotly.io.to_html` with additional javascript to improve usability.
 
+    Might become obsolete, when https://github.com/plotly/plotly.js/issues/998 gets
+    fixed.
+
     Injects `<script>`-tag with content from `to_html_patch.js` at the end of the HTML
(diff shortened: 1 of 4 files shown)


blackboxopt/visualizations/utils.py
```diff
@@ -171,3 +174,45 @@ def mask_pareto_efficient(costs: np.ndarray):
...
+    Returns:
+        Patched method.
+    """
+
+    @wraps(method)
+    def wrapped(*args, **kwargs):
+        html = method(*args, **kwargs)
+
+        # Test if title text contains "[BBO]"
+        if html.find('"title": {"text": "[BBO]') < 0:
```
dynobo, line 201: I introduced a convention, that the charts title has to start with `[BBO]`, otherwise the patch is not applied.

I did this to mitigate the conflicting goals of:
- Need to patch `plotly.io.to_html` globally, so our charts receive the patch, no matter how they are generated (`fig.show()` or `fig.to_html()` call `plotly.io.to_html()` internally)
- Minimize the side effects on the users, if they generates other charts with ploty.

What do you think? Any other ideas?

dynobo:
@sfalkner: Want to test drive this feature? I'd be curious about your feedback.

PR #33, opened 2022-01-05
add opt out of auto handling exceptions in optimization loops

LGro:

### DIFF
--- a/blackboxopt/optimization_loops/dask_distributed.py
+++ b/blackboxopt/optimization_loops/dask_distributed.py
@@ -37,10 +37,12 @@ class MinimalDaskScheduler:
         self,
         dask_client: dd.Client,
         objectives: List[Objective],
+        exit_on_unhandled_exception: bool,
         logger: logging.Logger,
     ):
         self.client = dask_client
         self.objectives = objectives
+        self.exit_on_unhandled_exception = exit_on_unhandled_exception
         self.logger = logger
         self._not_done_futures: Set = set()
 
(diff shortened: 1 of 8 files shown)


blackboxopt/optimization_loops/utils.py
```diff
@@ -55,7 +56,10 @@ def evaluation_function_wrapper(
     """
     try:
         evaluation = evaluation_function(evaluation_specification)
-    except Exception:
+    except Exception as e:
+        if exit_on_unhandled_exception:
+            raise e
```
dynobo, line 61: Just an idea: Could it be useful to have a custom exception, which the user can easier catch and work with? Something like:
```
    except Exception as e:
        if exit_on_unhandled_exception:
            raise BboEvaluationFunctionError from e
```

blackboxopt/optimization_loops/dask_distributed.py
```diff
@@ -102,6 +105,7 @@ def run_optimization_loop(
     dask_client: dd.Client,
     timeout_s: float = float("inf"),
     max_evaluations: int = None,
+    exit_on_unhandled_exception: bool = False,
```
dynobo, line 108: We really want `False` as default? At least I usually run earlier into the issue, that I miss the exceptions, where `True` would be helpful. If I remember correctly, this was also sometimes the case for the users. But I also don't know, how common those "numerical instabilities" are...

dynobo (commented):
Nice PR, really beneficial feature for the users!

dynobo (approved):

PR #34, opened 2022-01-10
add dict method to evaluation spec and evaluation

LGro:


blackboxopt/evaluation.py
```diff
@@ -106,6 +106,10 @@ def __len__(self):
     def to_json(self, **json_dump_kwargs):
         return json.dumps(asdict(self), **json_dump_kwargs)
 
+    # TODO: This could be an attribute, a method called dict or to_dict, what to go for?
+    def dict(self):
```
dynobo, line 110: I think my favorite would be `to_dict()`, like in `pandas.DataFrame`.

dynobo (approved):

blackboxopt/evaluation.py
LGro: I like the additional pandas convention argument, thanks :relaxed:

Issue #28, opened 2021-10-26
Make Visualizer objective aware


LGro:
Add support for objective (min/max directionality, objective names) in [Visualizer](https://github.com/boschresearch/blackboxopt/blob/main/blackboxopt/visualizations/visualizer.py#L179).

LGro:
Added test for the incumbent line calculation which is failing. Intending to double check my assumptions with @sfalkner.

PR #35, opened 2022-01-11
Feature/objectives in visualizer (#28)

LGro:
Given that we require an objective in the `Visualizer` now, this is breaking and would warrant another major version bump. What do you think @dynobo?

### DIFF
--- a/blackboxopt/visualizations/utils.py
+++ b/blackboxopt/visualizations/utils.py
@@ -10,7 +10,7 @@ from typing import List
 import numpy as np
 
 import blackboxopt
-from blackboxopt import Evaluation
+from blackboxopt import Evaluation, Objective
 
 
 def get_t0(evaluations: List[Evaluation]):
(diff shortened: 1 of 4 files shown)


dynobo:
> Given that we require an objective in the `Visualizer` now, this is breaking and would warrant another major version bump. What do you think @dynobo?
I don't have hard feelings on the core package regarding major bumps, so I'd be fine with it.

If we feel like we should lower the amount of major bumps, we might want to open a separate branch (`develop` or sth.) and move closer to full-blown gitflow, but I currently don't feel like this is necessary.

LGro:
> If we feel like we should lower the amount of major bumps, we might want to open a separate branch (`develop` or sth.) and move closer to full-blown gitflow, but I currently don't feel like this is necessary.

Which would only help if we would be ok with releasing less often but we usually want to release right away anyhow, so I'll bump the major version.

dynobo (approved):

PR #39, opened 2022-02-01
Feature/bohb ordinal parameters

LGro:
@sfalkner, this is an attempt to map ordinal parameters to integer parameters for all BOHB internal operations (sampling and iterations) despite that it would only be needed in the `BOHBSampler`, but I found the implementation proposed here to be the least invasive and easiest to read. Also, I wouldn't expect a negative effect on the iterations' racing performance since a numerical representation is used there anyhow. Let me know what you think :relaxed: 

FYI: @aeivazi, @dynobo

### DIFF
--- a/blackboxopt/optimizers/bohb.py
+++ b/blackboxopt/optimizers/bohb.py
@@ -12,10 +12,13 @@
 #     - docstrings and type hints
 
 import logging
+from copy import deepcopy
+from typing import Any, Dict, Tuple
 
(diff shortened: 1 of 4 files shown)


LGro:
@sfalkner will propose implementing support for ordinals via the ordered kernel in statsmodels, if this is successful, this PR will be deleted.

LGro:
Superseeded by #41

PR #40, opened 2022-02-03
make multi objective visualization objective direction aware

LGro:
The pareto front visualization was faulty when used with objective values where greater is better.

### DIFF
--- a/blackboxopt/visualizations/visualizer.py
+++ b/blackboxopt/visualizations/visualizer.py
@@ -5,7 +5,7 @@
 
 import datetime
 import itertools
-from typing import Dict, List, Tuple, Union
+from typing import Dict, List, Sequence, Tuple, Union
 
 import numpy as np
 import pandas as pd
@@ -15,6 +15,7 @@ import plotly.io._html
 import scipy.stats as sps
 
 from blackboxopt import Evaluation, Objective
+from blackboxopt.utils import get_loss_vector
 from blackboxopt.visualizations import utils
 
 QUALITATIVE_COLORS = px.colors.qualitative.G10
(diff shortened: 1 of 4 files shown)


blackboxopt/visualizations/visualizer.py
```diff
@@ -127,42 +128,56 @@ def create_hover_information(sections: dict) -> Tuple[str, List]:
...
-def multi_objective_visualization(evaluations: List[Evaluation]):
-    if not evaluations:
-        raise NoSuccessfulEvaluationsError
-
-    # Prepare dataframe for visualization
-    df = evaluations_to_df(evaluations)
+def _prepare_for_multi_objective_visualization(
+    df: pd.DataFrame, objectives: Sequence[Objective]
+) -> Tuple[pd.DataFrame, Dict[str, List[str]]]:
+    loss_matrix = np.array(
```
dynobo, line 134: What do you think about adding a comment here? Something along:

`# Unify potentially different objective directions (greater_is_better)`

blackboxopt/visualizations/visualizer.py
```diff
@@ -127,42 +128,56 @@ def create_hover_information(sections: dict) -> Tuple[str, List]:
...
     # Formatting for nicer output
     df["duration"] = df["duration"].round("s").astype("str")
     df["created_unixtime"] = df["created_unixtime"].dt.round("1s")
 
+    return df, hover_sections
+
+
+def multi_objective_visualization(
+    evaluations: List[Evaluation], objectives: Sequence[Objective]
+):
```
dynobo, line 167: I :heart: it, that we can now plot a subset of objectives, too. 

Just too bad, that `Evaluation` doesn't carry the objective's direction, otherwise we could have avoid the breaking change.

dynobo (approved):
Clean fix! :-)

PR #44, opened 2022-03-08
add option to override none replacement value in get loss vector util

LGro:

### DIFF
--- a/blackboxopt/utils.py
+++ b/blackboxopt/utils.py
@@ -3,7 +3,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-from typing import Dict, Optional, Sequence
+from typing import Dict, List, Optional, Sequence
 
 from blackboxopt.base import Objective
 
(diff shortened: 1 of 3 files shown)


blackboxopt/utils.py
```diff
@@ -3,29 +3,33 @@
...
     reported_objectives: Dict[str, Optional[float]],
-) -> list:
+    none_replacement: float = float("NaN"),
+) -> List[float]:
     """Convert reported objectives into a vector of known objectives.
 
     Args:
-        known_objectives: [description]
-        reported_objectives: [description]
+        known_objectives: A sequence of the objectives that e.g. an optimizer was
```
aeivazi, line 19: why `e.g.` here?
we could also add here: "Dictates the order for returned loss"

aeivazi (approved):

blackboxopt/utils.py
LGro: "e.g." because they could also be known for some other reason. And I'll happily add the return loss order detail.

PR #45, opened 2022-03-09
add post evaluation callback to sequential optimization loop

LGro:

### DIFF
--- a/blackboxopt/optimization_loops/sequential.py
+++ b/blackboxopt/optimization_loops/sequential.py
@@ -54,7 +54,7 @@ def run_optimization_loop(
             spurious errors due to e.g. numerical instability that should not halt the
             optimization loop.
         post_evaluation_callback: Reference to a callable that is invoked after each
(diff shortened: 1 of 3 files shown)


blackboxopt/optimization_loops/sequential.py
```diff
@@ -52,6 +53,8 @@ def run_optimization_loop(
             trace in the evaluation's `stacktrace` attribute. Set to True if there are
             spurious errors due to e.g. numerical instability that should not halt the
             optimization loop.
+        post_evaluation_callback: Reference to a callable that is invoked after each
+            evaluation.
```
sfalkner, line 57: Would it make sense if docstring here also reflects that the evaluation is the only argument to the callable? It is technically covered by the typehints, but maybe a half-sentence here wouldn't hurt?

sfalkner (approved):
Thanks @LGro for this. Looks good to me!

PR #46, opened 2022-03-17
fix evaluation accepts inf objective values

LGro:


sfalkner (approved):
Thanks @LGro for catching it and implementing this so quickly!

PR #48, opened 2022-04-07
fix mask pareto efficient visualization util

LGro:
We had the issue where points were considered pareto efficient when they were "on the pareto front" despite not being on a corner of the pareto front. This PR extends the test with two cases that failed before and are fixed now with what is hopefully also a more readable despite somewhat more complex implementation.


sfalkner (approved):
Thanks for spotting and fixing this!

PR #49, opened 2022-04-25
test on all supported python versions

LGro:
Given that we promise to support different Python versions, let's test that we actually do.
To that end this PR separates linting from testing and introduces a test matrix with Python versions 3.7, 3.8, and 3.9. We also explicitly exclude the yet unsupported 3.10 now in the pyproject.toml

FYI: @attilareiss and @sfalkner 

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -1,8 +1,8 @@
 name: ci-cd-pipeline
 
-on: [push, pull_request]
+on: push
 jobs:
-  test:
+  lint:
     runs-on: ubuntu-20.04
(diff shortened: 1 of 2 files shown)


aeivazi (approved):
Thank you @LGro!

dynobo:
Thanks for the initiative, @LGro :-)
In general, I prefer a setup where the tests can easily be done locally, too. In case of Python versions, `tox` might be the way to go. 
But testing just with gh actions is already a great improvement, too! :-)

dynobo (approved):

PR #50, opened 2022-05-02
Feature/update mkdocs dependencies

dynobo:
- Update mkdocs deps to (to avoid "jinja2 issue")
- Fix small issues that caused parts of the documentation being not parsed correctly

I bumped the patch version, so we can rebuild a (fixed) version of the documentation easily. We could also wait until the next "real" release that changes more than the docs, but I thought it also can't hurt to drive out our release pipeline from time to time, so we see if anything breaks.


pyproject.toml
```diff
@@ -26,21 +26,21 @@ distributed = {version = "^2021.2.0", optional = true}
 pandas = {version = "^1.2.4", optional = true}
 
 [tool.poetry.dev-dependencies]
-black = "^21.8b0"
-isort = "^5.7.0"
-pytest = "^6.2.2"
-mypy = "^0.910"
-pytest-timeout = "^1.4.2"
-pytest-cov = "^2.11.1"
+black = "^22.3.0"
```
LGro, line 29: We're using a pre-commit hook for the exact version `22.3.0` but here don't fix it to `black = "22.3.0"` but instead allow variation in the minor version, is that intended?

LGro:
Thanks a lot @dynobo
:rocket:

LGro (approved):

pyproject.toml
dynobo: Good catch, @LGro. But before fixing `black` in the pyproject.toml, I'd like to investigate the use external commit hooks in our pre-commit-config.yaml:

Those we introduced by a template we used, but so far I only see downside of using external hooks against just running the tools directly from poetry environment (`poetry run black ...`):
- Versions have to be maintained at two places
- (Some) options have to be maintained at two place
- Tools get installed twice
- Less control over what is actually installed from that remote repo

I'd propose to merge this PR now and follow up on the pre-commit topic in a different task. What do you think?

LGro: Love the proposal, I'm all for ditching the external pre-commit hooks that run in separate environments because all we need is anyhow always installed.

dynobo: @LGro, did the change in the latest commits, but decided to keep the generic hooks (line-endings, large file, etc.) from the remote repo, as I think they are still valuable. Do you agree?

PR #52, opened 2022-06-27
Add parallel coordinates plot

dynobo:
Did that during my train ride as one way to introspect the evaluation results in a higher dimensional search space with a [parallel coordinates plot](https://plotly.com/python/parallel-coordinates-plot/).

Unfortunately, I can't upload an image here (proxy?), but you can debug into the the unit test and and do a `fig.show()` there, which also gives a better impression about the interactivity.

Known issues we might want to tackle in the future:
- Ordering of ordinal parameter values is not ensured
- Axis of Integer parameters might still show axis ticks on fractional positions

### DIFF
--- a/blackboxopt/visualizations/visualizer.py
+++ b/blackboxopt/visualizations/visualizer.py
@@ -5,7 +5,7 @@
 
 import datetime
 import itertools
-from typing import Dict, List, Optional, Sequence, Tuple, Union
+from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
 
 import numpy as np
 import pandas as pd
@@ -196,8 +196,30 @@ def multi_objective_visualization(
 
 
 def parallel_coordinate_plot_parameters(
-    evaluations: List[Evaluation], color_by: Optional[Objective] = None
+    evaluations: List[Evaluation],
+    columns: Optional[List[str]] = None,
+    color_by: Optional[str] = None,
 ):
+    """Create an interactive parallel coordinate plot.
+
+    Useful to investigate relationships in a higher dimensional search space and the
+        optimization's objective(s).
+
+    Args:
+        evaluations: Evaluations to plot.
+        columns: Names of columns to show. Can consist of parameter names and/or
+            objective names. If `None`, all parameters and all objectives are displayed.
+        color_by: Parameter name or objective name. The corresponding column will be
+            shown at the very right, it's value will be used for the color scale. If
+            `None`, all lines have the same color.
+
+    Returns:
+        Plotly figure
+
+    Raised:
+        NoSuccessfulEvaluationsError: In case `evaluations` does not contain at least
+            one successful evaluation (an evaluation with objective value != `None`).
+    """
     if not evaluations:
         raise NoSuccessfulEvaluationsError
 
@@ -206,17 +228,25 @@ def parallel_coordinate_plot_parameters(
 
     # Drop unused columns and indices
     df = df[["configuration", "objectives"]]
-    parameter_cols = df["configuration"].columns.to_list()
     objective_cols = df["objectives"].columns.to_list()
-    color_by_col = color_by.name if color_by else objective_cols[0]
-
     df = df.droplevel(0, axis=1)
 
-    # Prepare a coordinate (vertical line) for every parameter
+    # If no columns are specified, use all:
+    if not columns:
+        columns = df.columns.to_list()
+
+    # Prepare a coordinate (vertical line) for every column
     coordinates = []
-    for column in parameter_cols:
+    colored_coordinate = {}
+    for column in columns:
+        coordinate: Dict[str, Any] = {}
+
+        if column in objective_cols:
+            coordinate["label"] = f"<b>Objective: {column}</b>"
+        else:
+            coordinate["label"] = column
+
         parameter_type = df[column].dtype.name
-        coordinate = dict(label=column)
         if parameter_type.startswith("float") or parameter_type.startswith("int"):
             # Handling floats and integers as same, vecause unfortunately it's hard to
             # use integers only for ticks and still be robust regarding a large range
@@ -238,23 +268,21 @@ def parallel_coordinate_plot_parameters(
             raise NotImplementedError(
                 f"Unknown column type: {column}<{parameter_type}>"
             )
-        coordinates.append(coordinate)
+        if column == color_by:
+            colored_coordinate = coordinate
+        else:
+            coordinates.append(coordinate)
 
-    for column in objective_cols:
-        if column == color_by_col:
-            continue
-        coordinate = dict(label=f"<b>🎯 {column}</b>", values=df[column])
-        coordinates.append(coordinate)
-
-    # Add the color_by-objective as last scale
-    coordinates.append(dict(label=f"<b>🎯 {color_by_col}</b>", values=df[color_by_col]))
+    # Append colored coordinate to the end (right)
+    if colored_coordinate:
+        coordinates.append(colored_coordinate)
 
     # Plot
     return go.Figure(
         data=go.Parcoords(
             line=dict(
                 # Color lines by objective value
-                color=df[color_by_col],
+                color=df[color_by] if color_by else None,
                 colorscale=px.colors.diverging.Tealrose,
                 showscale=True,
                 # Use colorbar as kind of colored extension to the axis
@@ -268,9 +296,7 @@ def parallel_coordinate_plot_parameters(
             ),
             dimensions=coordinates,
         ),
-        layout=dict(
-            title=f"[BBO] Parallel coordinates plot colored by '{color_by_col}'-value"
-        ),
+        layout=dict(title=f"[BBO] Parallel coordinates plot"),
     )
 
 
--- a/blackboxopt/visualizations/visualizer.py
+++ b/blackboxopt/visualizations/visualizer.py
@@ -6,6 +6,7 @@
 import datetime
 import itertools
 import warnings
+from collections import Counter
 from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
 
 import numpy as np
@@ -237,6 +238,14 @@ def parallel_coordinate_plot_parameters(
     # If no columns are specified, use all:
     if not columns:
         columns = df.columns.to_list()
+
+    ambigious_columns = [k for k, v in Counter(df[columns].columns).items() if v > 1]
+    if ambigious_columns:
+        raise ValueError(
+            "All columns to plot must have a unique name, but those are ambigious: "
+            + f"{ambigious_columns}. Either rename parameters/settings/objective to "
+            + "be unique or provide only the unambigious ones as `columns` argument."
+        )
 
     # Prepare a coordinate (vertical line) for every column
     coordinates = []
--- a/blackboxopt/visualizations/visualizer.py
+++ b/blackboxopt/visualizations/visualizer.py
@@ -5,6 +5,7 @@
 
 import datetime
 import itertools
+import warnings
 from typing import Any, Dict, List, Optional, Sequence, Tuple, Union
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


LGro:
Thanks @dynobo, that's such a helpful addition :relaxed: 

Concerning the multi objective scenario, if we want to support it the way that this PR proposes with picking one objective to color the configurations, how do you feel about adding the other objectives at least as another column so that one can use the nice interactive filter to get the subset one is interested in?

dynobo:
> how do you feel about adding the other objectives at least as another column so that one can use the nice interactive filter to get the subset one is interested in?

Interesting idea, I haven't thought about that! My plan was, that a user just generates multiple plots, one for each objective, but I agree: having all in one chart could also be beneficial. 

I'm going to try it and see how it looks like.

dynobo:
@LGro, the last commit plots all objectives now. Accordingly, I renamed the argument to `color_by` and made it optional (with defaulting to first objective). Would you mind taking another look? What do you think about the 🎯 to mark objectives? :-)

blackboxopt/visualizations/visualizer.py
```diff
@@ -195,6 +195,85 @@ def multi_objective_visualization(
...
+        raise NoSuccessfulEvaluationsError
+
+    # Prepare dataframe for visualization
+    df = evaluations_to_df(evaluations)
+
+    # Drop unused columns and indices
+    df = df[["configuration", "objectives"]]
+    parameter_cols = df["configuration"].columns.to_list()
+    objective_cols = df["objectives"].columns.to_list()
+    color_by_col = color_by.name if color_by else objective_cols[0]
```
LGro, line 211: Since there is no necessarily unified order in the evaluations' objective values dictionaries I would consider taking the "first" as the default to be too arbitrary and would rather always require an explicit objective to color by. But then again not having to specify it is nice for the single objective cases, so let's leave it as is. And it also feels weird to only require it for the multi objective case (i.e. raise value error if multiple objective values and `color_by=None`, or what do you think?

LGro:
> Accordingly, I renamed the argument to `color_by` and made it optional (with defaulting to first objective). Would you mind taking another look? What do you think about the dart to mark objectives? :-)

Love the argument renaming and while I was skeptical just reading about the emoji as an objective indicator, visually it works surprisingly well. However, I'm still thinking it might pay to have that explained somewhere in a legend of sorts or to just use `Objective↑: ` as a prefix for greater is better objectives and `Objective↓: ` for the other direction. Let me know what you think.

blackboxopt/visualizations/visualizer.py
dynobo: Honestly, after fiddling around a bit, I'm not convinced anymore that showing multiple objectives in one chart is a good idea, and I start preferring the initial solution more and more:

1. **Chart type:** To me it feels like a parallel coordinates plot is not suitable to plot an additionally quality (type: parameter vs. objective). The main reason is, that it relies on the coloring as key: The color is the thing that connects all coordinates. We can only color one coordinate, so there will never be a connection between a non-colored-objective and the parameters, (except for the ones directly next to it, where the lines give an indication). Such a plot would mainly allow comparing the colored objective to each individual other objective. For that purpose, we have the more suitable pareto-plot. Comparing the the non-colored objectives to the parameters will always be hard/cumbersome in such a chart.
2. **Developer UX:** If we allow plotting only one explicitly required objective, everything is totally transparent. As soon as we want to support multiple objectives, we'd have to make assumptions or handle different cases.
3. **Visual:** The colored objective is visually separated from the parameters. As soon as we add more objectives, we kind of start putting apples and oranges on the same visual level, putting an indicator to the label-name is just a mediocre workaround.

Another options, that would leave the decision to user, 
Require `objective:Union[Objective, List[Objective]], color_by: Optional[Objective]=None` and only plot the objectives provided. In case a list of objectives is provided but no `color_by` is set, the first objective in the list will be used for coloring. 

For the multi objective case, my dream chart would be multi-row, with one paracord chart per row for each objective, but with the ability to crossfilter (if you select some evaluations in one of the rows, they are also selected in the others)). Unfortunately, this is time consuming to implement. I could try to approximate this by returning a multi-row chart without crossfilter, but that might need some time. It would certainly have the benefit, that the `color_by`/`objective` arg would become obsolete.

What do you think about those thoughts?

LGro: I feel you, so here is a proposal that should be flexible enough to not artificially restrict use cases with multiple objectives while also allowing for a smooth UX in the simple single objective case while not causing us devs much pain if complicated case handling. But before I go there I just want to highlight that there is value in a version of this plot without colors because one still gets the option to subselect and see a vague notion of "density". That being said, here is the proposal:

```python
def parallel_coordinate_plot_parameters(
    evaluations: List[evaluations],
    columns: Optional[List[str]] = None,
    color_by: Optional[str] = None
)
```

Where columns can be None for all available or a specified subset of the columns (config keys, settings keys, objective keys). And we only color the plot if `color_by` is set, and it can be any column of the data.

This allows for the most prominent use cases I currently see...

1) simple single objective case
```python
parallel_coordinate_plot_parameters(evals, color_by: "loss")
```

2) comparing only objectives for moo case without color
```python
parallel_coordinate_plot_parameters(evals, columns=["loss", "params", "flops"])
```

3) having few objectives in one and optionally use one to color
```python
parallel_coordinate_plot_parameters(evals) # or
parallel_coordinate_plot_parameters(evals, color_by="chosen objective")
```

LGro: **RE: column**
Yes, documentation - also, I am hoping that it helps for people to see the unrestricted state, because there they see the actual vertical lines i.e. columns for this to additionally drive home the point what columns are conceptually.
Also, please include settings keys if they aren't already (most prominently fidelity), because that's something to interactively filter by for sure.

**RE: cumbersome use case**
Could you elaborate on the insights you derive from a visualization of multi objective run results where parameters and only one objective are shown? Usually, I find one does multi objective optimization because one is interested in a trade off - i.e. looking at the combination of the objectives, as opposed to dropping the information about all but one objective. Can also be a direct conversation separate from this PR, since it's not blocking it.

LGro: Thanks for the trust and the discussion - I hope it'll be rewarded with a good experience down the line :relaxed:

dynobo: Could you please review my latest commit which implements your proposal? Especially the docstring... :innocent:

dynobo: >  Also, please include settings keys if they aren't already (most prominently fidelity), because that's something to interactively filter by for sure.

Ah, good idea! I'll do that tomorrow :-D

sfalkner: Just to add another data point for the more flexible solution, which I like very much.
In the AIMO context, we saw what our colleagues would look at for MOO results, and they would predominantly look at a parallel coordinate plot with only the objectives. They didn't want to see the actual configuration (it was too high dimensional and the parameters where hard to interpret), but wanted a single plot to see and compare different trade-offs. This would now also be possible with the latest change :)

blackboxopt/visualizations/visualizer.py
```diff
@@ -195,6 +196,117 @@ def multi_objective_visualization(
...
+
+    Useful to investigate relationships in a higher dimensional search space and the
+        optimization's objective(s).
+
+    Args:
+        evaluations: Evaluations to plot.
+        columns: Names of columns to show. Can contain parameter names, objective names
+            and settings keys. If `None`, all parameters, objectives and settings are
+            displayed.
+        color_by: Parameter name, objective name, settings key. The corresponding column
```
sfalkner, line 214: Reading this, it is not clear to me if the value for `color_by` needs to already be in `columns` or whether it will be plotted twice then. I can figure this out by looking at the code, but from the docstring, this remains unclear.

blackboxopt/visualizations/visualizer.py
```diff
@@ -195,6 +196,117 @@ def multi_objective_visualization(
...
+        if column in objective_cols:
+            coordinate["label"] = f"<b>Objective: {column}</b>"
+        elif column in settings_cols:
+            coordinate["label"] = f"Setting: {column}"
+        else:
+            coordinate["label"] = column
+
+        parameter_type = df[column].dtype.name
+        if parameter_type.startswith("float") or parameter_type.startswith("int"):
+            # Handling floats and integers as same, vecause unfortunately it's hard to
```
sfalkner, line 256:
```suggestion
            # Handling floats and integers the same, because unfortunately it's hard to
```

blackboxopt/visualizations/visualizer.py
```diff
@@ -195,6 +196,117 @@ def multi_objective_visualization(
...
+            # use integers only for ticks and still be robust regarding a large range
+            # of values.
+            coordinate["values"] = df[column]
+        elif parameter_type in ["object", "bool"]:
+            # Encode categorical values to integers. Unfortunately, ordinal parameters
+            # loose there ordering, as there is no information about the order in the
```
sfalkner, line 262: If we want to keep the ordering, we would need to also have the search space here. We can add this later, if we feel that this becomes more important.

sfalkner (approved):
Thanks @dynobo for this flexible and clean implementation!

blackboxopt/visualizations/visualizer.py
LGro: +1 for doing that later when we needed

blackboxopt/visualizations/visualizer.py
dynobo: Currently it behaves in the most flexible way, meaning `columns=["a","b","c"], color_by="d"`will show only three columns (a, b, c), but they will be colored by the value of d. 

I don't know if this is of any use, and also in this case the visualization is a bit misleading with having the color-bar according to d right next to the most right column c, that's why I didn't want to actually "promote" this way of usage by mentioning it explicitly that there is no such restriction.

What do you think about that? Should we enforce that `color_by` is included in the displayed `columns`? Should we make it more explicit in the docstring, that there is no such restriction?

blackboxopt/visualizations/visualizer.py
dynobo: Implemented.

dynobo:
@sfalkner , @LGro: Note about the current "flexible" implementation: by allowing to use the names of parameters, objectives and settings in a single list, we are prone to naming collisions. In case we have e.g. a setting with the same name as an parameter, the currently implementation will not hold. But solving this most likely is a larger effort, and should probably done in a separate PR/issue.

LGro:
> @sfalkner , @LGro: Note about the current "flexible" implementation: by allowing to use the names of parameters, objectives and settings in a single list, we are prone to naming collisions. In case we have e.g. a setting with the same name as an parameter, the currently implementation will not hold. But solving this most likely is a larger effort, and should probably done in a separate PR/issue.

Indeed, how about we raise a user warning for now?

dynobo:
> Indeed, how about when this happens we raise a user warning for now?

This alone won't help: The implementation will raise an exception with duplicate column names, because they are used as identifiers, for the transformations as well as for the plots. One would have to implement a smart renaming of the columns that still works with inputting the original column names (which can be ambiguous).

I see two options:
1. Raise a meaningful ValueError early asking to provide un-ambiguous columns only
2. Warn and drop duplicate columns (dropping prio: settings, parameters, objectives?)

LGro:
> I see two options:
> 
>     1. Raise a meaningful ValueError early asking to provide un-ambiguous columns only
> 
>     2. Warn and drop duplicate columns (dropping prio: settings, parameters, objectives?)

Both fine, since I'd expect in both cases the user would act and manually rename, so what ever is easier to implement.

LGro:
> EDIT: Hang on, we could also prefix them with C: for config keys, S: for settings keys and O: for objectives keys to avoid collisions and help keep track of where columns in the final plot came from - that's my preferred solution, I think.

With that we wouldn't need to raise on duplicates (0edc7d5) because nothing can collide anymore, or what do you think?

LGro (approved):

PR #54, opened 2022-06-29
Feature/default test for fixed parameters

LGro:
We noticed for custom optimizer implementations that it could make sense to ensure that they respect fixed parameters, so this PR introduces a reference test that helps spot deviations from that expected behaviour.

I also added missing type hints and fixed docstring typos in the reference tests - if you'd like to de-clutter the diff, consider looking at the individual commits one by one, they are almost nicely curated (the one to spot the odd one out gets a drink :smile:).

### DIFF
--- a/blackboxopt/optimizers/testing.py
+++ b/blackboxopt/optimizers/testing.py
@@ -44,7 +44,11 @@ def _initialize_optimizer(
 
 
 def optimize_single_parameter_sequentially_for_n_max_evaluations(
-    optimizer_class, optimizer_kwargs: dict, n_max_evaluations: int = 20
+    optimizer_class: Union[
+        Type[SingleObjectiveOptimizer], Type[MultiObjectiveOptimizer]
+    ],
+    optimizer_kwargs: dict,
+    n_max_evaluations: int = 20,
 ) -> bool:
     """[summary]
 
@@ -101,7 +105,12 @@ def optimize_single_parameter_sequentially_for_n_max_evaluations(
     return True
 
 
-def is_deterministic_with_fixed_seed(optimizer_class, optimizer_kwargs: dict) -> bool:
+def is_deterministic_with_fixed_seed(
+    optimizer_class: Union[
+        Type[SingleObjectiveOptimizer], Type[MultiObjectiveOptimizer]
+    ],
+    optimizer_kwargs: dict,
+) -> bool:
     """Check if optimizer is deterministic.
 
     Repeatedly initialize the optimizer with the same parameter space and a fixed seed,
@@ -111,7 +120,7 @@ def is_deterministic_with_fixed_seed(optimizer_class, optimizer_kwargs: dict) ->
 
     Args:
         optimizer_class: Optimizer to test.
-        optimizer_kwargs: Expected to contain additional arguments for initializating
+        optimizer_kwargs: Expected to contain additional arguments for initializing
             the optimizer. (`search_space` and `objective(s)` are set automatically
             by the test.)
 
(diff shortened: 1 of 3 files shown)


LGro:
I'll squash and sign-off to comply with DCO after the review.

blackboxopt/optimizers/testing.py
```diff
@@ -218,9 +237,48 @@ def raises_evaluation_error_when_reporting_unknown_objective(
...
+    Returns:
+        `True` if the test is passed.
+    """
+    space = ps.ParameterSpace()
+    space.add(ps.ContinuousParameter("my_fixed_param", (-10.0, 200.0)))
+
+    fixed_value = 1.0
+    space.fix(my_fixed_param=1.0)
+
+    for _ in range(5):
```
lreeb, line 264: Just out of curiosity, why repeat this 5 times? Should it not be deterministic?

blackboxopt/optimizers/testing.py
```diff
@@ -218,9 +237,48 @@ def raises_evaluation_error_when_reporting_unknown_objective(
...
+            by the test.)
+
+    Returns:
+        `True` if the test is passed.
+    """
+    space = ps.ParameterSpace()
+    space.add(ps.ContinuousParameter("my_fixed_param", (-10.0, 200.0)))
+
+    fixed_value = 1.0
+    space.fix(my_fixed_param=1.0)
```
lreeb, line 262: Since you defined `fixed_value`, might be good to use it here instead of the literal `1.0`

lreeb (approved):
That was fast :rocket:  Thanks @LGro!

blackboxopt/optimizers/testing.py
LGro: There is a small chance that a non-fixed parameter would sample exactly 1.0, and while 5 is certainly arbitrary, I'd still vote for trying more than once.

blackboxopt/optimizers/testing.py
LGro: Guess who reverted back and forth too much - that was the idea, thanks @lreeb!

sfalkner:
Was there a specific incident that made you think that this might be broken or is this purely a safety measure?
Ideally, the optimizers don't see any fixed parameters, if they rely on the numerical representation. Everything should be handled in `ParameterSpace`.

blackboxopt/optimizers/testing.py
```diff
@@ -218,9 +237,48 @@ def raises_evaluation_error_when_reporting_unknown_objective(
...
+        optimizer_class: Optimizer to test.
+        optimizer_kwargs: Expected to contain additional arguments for initializing
+            the optimizer. (`search_space` and `objective(s)` are set automatically
+            by the test.)
+
+    Returns:
+        `True` if the test is passed.
+    """
+    space = ps.ParameterSpace()
+    space.add(ps.ContinuousParameter("my_fixed_param", (-10.0, 200.0)))
```
sfalkner, line 259: This `ParameterSpace` has actually no changeable parameter anymore, which could lead to all kind of problems for the optimizers. Is this intended, or would it make sense to add another dummy parameter?

blackboxopt/optimizers/testing.py
```diff
@@ -218,9 +237,48 @@ def raises_evaluation_error_when_reporting_unknown_objective(
...
+    for _ in range(5):
+        opt = _initialize_optimizer(
+            optimizer_class,
+            optimizer_kwargs,
+            objective=Objective("loss", False),
+            objectives=[Objective("loss", False)],
+            space=space,
+        )
+        es = opt.generate_evaluation_specification()
+        assert es.configuration["my_fixed_param"] == fixed_value
```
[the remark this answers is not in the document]
LGro: No, since in each iteration the optimizer is created from scratch. Would the optimizer be instantiated outside of the loop, your observation would be accurate.

blackboxopt/optimizers/testing.py
LGro: Depends a bit on how much we rely on users doing reasonable things. Feeding an all fixed space into an optimizer doesn't make too much sense. So proposal: We add a dummy parameter here and add a different test that checks that an all fixed space or a space without parameters raises a ValueError when given to any optimizer - what do you think?

blackboxopt/optimizers/testing.py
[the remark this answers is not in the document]
sfalkner: You are absolutely right, but then I have to ask whether we really want to test 5 times if the first proposed eval_spec contains the fixed parameter. Most model-based optimizers would start with a random sample that might come directly from `ParameterSpace.sample()`, rather than building a model. I would argue that it would be more interesting to see if 5 evaluations can be queried if results are reported back. That way, any non-trivial optimizer will be forced to do its non-trivial thing, which is what we probably want to test.

blackboxopt/optimizers/testing.py
sfalkner: That's one option, if we want to cover that. I was mostly concerned about testing with a degenerate case where `len(space) ==0`, which can lead to empty arrays and other nasty bugs that are unrelated to what this test wants to achieve.
In the case that we want to prevent users to accidentally optimizing zero parameters, I agree that the optimizers should throw an error, but I have no strong opinion if that level of sanity checks is required here.

blackboxopt/optimizers/testing.py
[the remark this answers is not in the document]
LGro: Thanks, that makes a lot of sense. I changed it accordingly.

sfalkner (approved):
Thanks @lreeb and @LGro for adding this!

sfalkner (approved):
Thanks @lreeb and @LGro for adding this!

PR #56, opened 2022-06-29
add support for Python 3.10

LGro:
Staying up to date :relaxed:


dynobo (approved):
Wooha! Only one more minor to [3.11](https://docs.python.org/3.11/whatsnew/3.11.html) with its 10-60% speed improvements...

PR #58, opened 2022-08-08
Fix parallel coordinate plot breaking in case of non-hashable parameter values

dynobo:
I notice that issue, when I tried to visualize the run with a parameterspace containing:

```python
ps.CategoricalParameter("p1", (None, [0,1,2], [3,2,4]))
```

While this also can be avoided on user side by choosing tuples instead of list, the proposed change of converting to string first makes the visualization robust for such cases without any downsides (I think?).

### DIFF
--- a/blackboxopt/__init__.py
+++ b/blackboxopt/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "4.4.4"
+__version__ = "4.4.5"
 
 from parameterspace import ParameterSpace
 
--- a/blackboxopt/visualizations/visualizer.py
+++ b/blackboxopt/visualizations/visualizer.py
@@ -280,7 +280,9 @@ def parallel_coordinate_plot_parameters(
             # Encode categorical values to integers. Unfortunately, ordinal parameters
             # loose there ordering, as there is no information about the order in the
             # evaluations.
-            df[column] = df[column].astype("category")
+            # The string conversion is necessary for unhashable entries, e.g. of
+            # type List, which can't be casted to categories.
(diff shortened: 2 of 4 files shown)


LGro:
Thanks for catching this and proposing such a simple fix.
Could we run into issues where values are somehow complex without guaranteed internal order so that the same value with a different order would lead to different string representations? Because then they would be represented separately in the visualization even though they are equal.

LGro (approved):

dynobo:
> Could we run into issues where values are somehow complex without guaranteed internal order so that the same value with a different order would lead to different string representations?

I guess not? The `configuration` in the `Evaluation` / `EvaluationSpecification` is always generated by us from the `ParameterSpace`, so I do not see how the order could change inconsistently. 

(Of course, if a users start messing with the `configuration`-properties on there own, this might happen, or if someone passes a references to e.g. a list as parameter value and starts manipulating it, but that's nothing I would try to cover.)

LGro:
> I guess not? The `configuration` in the `Evaluation` / `EvaluationSpecification` is always generated by us from the `ParameterSpace`, so I do not see how the order could change inconsistently.

Folks don't need to use our visualization exclusively for evaluations from BBO runs, though. We're just specifying a common format for evaluations and provide visualization tools. In case a user starts out with manual evaluations or evaluations from a different system (e.g. converting from evaluations of ConfigSpace based search spaces) as initial data, the same order for e.g. values of type dict isn't guaranteed.
We can still decide to accept this risk, if the complexity of handling that edge case brings too much overhead to the implementation. I'd mainly like us to make that choice consciously.

PR #59, opened 2022-08-22
Upgrade plotly to new major version

dynobo:
Catching up to the latest major version. There is still no feature to pin tooltips, that's why I also had to update our small hack to handle that.

FYI: Here's what's new in [plotly.py](https://github.com/plotly/plotly.py/blob/master/CHANGELOG.md) and [plotly.js](https://github.com/plotly/plotly.js/blob/master/CHANGELOG.md).


sfalkner (approved):
Thanks @dynobo for keeping things up-to-date!

PR #60, opened 2022-08-31
display single scatter plot instead of scatter matrix for two objectives

LGro:
Before this PR, when visualizing pairwise objective scatter plots with two objectives, we showed the same scatter plot matrix like with more than two objectives even though it is not necessary. This PR now shows a simple scatter plot in the two objectives case to save space and be less confusing visually due to displaying the same plot twice, mirrored, in a matrix diagonal.


dynobo (approved):
Thanks! :rocket:

PR #61, opened 2022-09-06
add filter pareto front from eval util

LGro:
The time has come and we benefit from a function that just takes a list of evaluations and filters out the pareto front, in addition to / building on the previous implementation of the loss matrix based mask.

### DIFF
--- a/blackboxopt/visualizations/utils.py
+++ b/blackboxopt/visualizations/utils.py
@@ -12,6 +12,9 @@ import numpy as np
 import blackboxopt
 from blackboxopt import Evaluation, Objective
 
+# For backwards compatibility reasons
+from blackboxopt.utils import mask_pareto_efficient
+
 
 def get_t0(evaluations: List[Evaluation]):
     return min([e.created_unixtime for e in evaluations])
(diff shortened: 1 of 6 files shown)


tests/utils_test.py
```diff
@@ -42,3 +46,46 @@ def test_get_loss_vector_with_custom_none_replacement():
...
+        Evaluation(configuration={"i": 4}, objectives={"loss": 3.1, "score": -1.1}),
+        Evaluation(configuration={"i": 5}, objectives={"loss": 0.1, "score": -1.0}),
+        Evaluation(configuration={"i": 6}, objectives={"loss": 0.0, "score": -1.1}),
+        Evaluation(configuration={"i": 7}, objectives={"loss": -1.0, "score": -2.0}),
+    ]
+    pareto_efficient = filter_pareto_efficient(
+        evals, [Objective("loss", False), Objective("score", True)]
+    )
+    assert len(pareto_efficient) == 4
+    assert set([0, 2, 3, 7]) == set([e.configuration["i"] for e in pareto_efficient])
```
LGro, line 91: This is now duplicating test logic from `mask_pareto_efficient` more than it would need. Effectively, we just need to test here that the objective inversion and filtering of the list work. However, I found the overhead low enough and the benefit large enough that we get many edge cases here on this level as well to accept the duplication. Let me know what you think about this aspect.

blackboxopt/visualizations/utils.py
```diff
@@ -172,27 +175,6 @@ def add_plotly_buttons_for_logscale(fig):
...
-    for i, c in enumerate(costs):
-        if not is_efficient[i]:
-            continue
-
-        # Keep any point with a lower cost or when they are the same
-        efficient = np.any(costs[is_efficient] < c, axis=1)
-        duplicates = np.all(costs[is_efficient] == c, axis=1)
-        is_efficient[is_efficient] = np.logical_or(efficient, duplicates)
-
-    return is_efficient
```
LGro, line 193: This just moved to the general, visualization unspecific, utils (so did the test).

tests/utils_test.py
dynobo: Agree, it's worth it!

PR #63, opened 2022-10-07
add conditional space reference test

LGro:
This PR adds a test to the default compatibility test suite that contains a condition in the search space, i.e. has a hierarchical search space. This change requires all optimizers that test against the default tests to support conditional spaces.


aeivazi (approved):
Great test, very useful to have :) Thanks @LGro

PR #64, opened 2022-10-18
Add additional reference test for determinism

dynobo:
Proposal for a test to verify deterministic behavior when reporting a list of evaluations at once (here as initial data), even if the order of the list varies. 

Currently all optimizer pass the test (**EDIT:** except BOHB), but still I'm not 100% sure: is this _always_ the _desired_ behavior? Or are there cases, where we expect and desire different optimizer behavior depending on the order?

### DIFF
--- a/blackboxopt/optimizers/testing.py
+++ b/blackboxopt/optimizers/testing.py
@@ -51,7 +51,7 @@ def optimize_single_parameter_sequentially_for_n_max_evaluations(
     ],
     optimizer_kwargs: dict,
     n_max_evaluations: int = 20,
-) -> bool:
+):
     """[summary]
 
     Args:
@@ -104,15 +104,13 @@ def optimize_single_parameter_sequentially_for_n_max_evaluations(
         evaluation = eval_spec.create_evaluation(objectives=evaluation_result)
         optimizer.report(evaluation)
 
-    return True
-
 
 def is_deterministic_with_fixed_seed_and_larger_space(
     optimizer_class: Union[
         Type[SingleObjectiveOptimizer], Type[MultiObjectiveOptimizer]
     ],
     optimizer_kwargs: dict,
-) -> bool:
+):
     """Check if optimizer is deterministic.
 
     Repeatedly initialize the optimizer with the same parameter space and a fixed seed,
@@ -150,7 +148,6 @@ def is_deterministic_with_fixed_seed_and_larger_space(
         final_configurations.append(es2.configuration.copy())
 
     assert final_configurations[0] == final_configurations[1]
-    return True
 
 
 def is_deterministic_when_reporting_shuffled_evaluations(
@@ -158,7 +155,7 @@ def is_deterministic_when_reporting_shuffled_evaluations(
         Type[SingleObjectiveOptimizer], Type[MultiObjectiveOptimizer]
     ],
     optimizer_kwargs: dict,
-) -> bool:
+):
     """Check if determinism isn't affected by the order of initially reported data.
 
     Repeatedly initialize the optimizer with the same parameter space and a fixed seed.
@@ -228,15 +225,13 @@ def is_deterministic_when_reporting_shuffled_evaluations(
     assert initial_configs_run_0 != initial_configs_run_1
     assert configs_run_0 == configs_run_1
 
-    return True
-
 
 def handles_reporting_evaluations_list(
     optimizer_class: Union[
         Type[SingleObjectiveOptimizer], Type[MultiObjectiveOptimizer]
     ],
     optimizer_kwargs: dict,
-) -> bool:
+):
     """Check if optimizer's report method can process an iterable of evaluations.
 
     All optimizers should be able to allow reporting batches of evaluations. It's up to
@@ -266,7 +261,6 @@ def handles_reporting_evaluations_list(
         evaluations.append(evaluation)
 
     opt.report(evaluations)
-    return True
 
 
 def raises_evaluation_error_when_reporting_unknown_objective(
@@ -274,7 +268,7 @@ def raises_evaluation_error_when_reporting_unknown_objective(
         Type[SingleObjectiveOptimizer], Type[MultiObjectiveOptimizer]
     ],
     optimizer_kwargs: dict,
-) -> bool:
+):
     """Check if optimizer's report method raises exception in case objective is unknown.
 
     Also make sure that the faulty evaluations (and only those) are included in the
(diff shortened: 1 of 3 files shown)


dynobo:
Note: We might want to integrate the fix for failing BOHB in the same PR.

blackboxopt/optimizers/testing.py
```diff
@@ -148,6 +150,70 @@ def is_deterministic_with_fixed_seed(
...
+    optimizing and check if the generated configurations for all optimizers are equal.
+
+    Args:
+        optimizer_class: Optimizer to test.
+        optimizer_kwargs: Expected to contain additional arguments for initializing
+            the optimizer. (`search_space` and `objective(s)` are set automatically
+            by the test.)
+
+    Returns:
+        `True` if the test is passed.
```
sfalkner, line 172: Usually, tests fail on a problem but "quietly return"  when they pass. Is this considered "best practice" for reusable tests?

blackboxopt/optimizers/testing.py
```diff
@@ -148,6 +150,70 @@ def is_deterministic_with_fixed_seed(
...
+        return float(np.squeeze(y))
+
+    optimization_runs: Dict[int, list] = {0: [], 1: []}
+    for run_idx in optimization_runs:
+        opt = _initialize_optimizer(
+            optimizer_class,
+            optimizer_kwargs,
+            objective=Objective("loss", False),
+            objectives=[Objective("loss", False)],
+            space=space,
```
sfalkner, line 192: Would it makes sense to set the `seed` argument here explicitly to make it obvious that the underlying rng is fixed? I know that it is fixed by default in `_initialize_optimizer`, but I actually needed to go to that function and check to convince myself that the seed is fixed.

sfalkner (approved):
Thanks, @dynobo for adding this test!

blackboxopt/optimizers/testing.py
dynobo: Good question, @sfalkner, I just followed the pattern of the existing reference tests for consistency, but was also a bit confused by it. I'd suggest that someone removes those `return`s from all reference tests in separate PR.

blackboxopt/optimizers/testing.py
dynobo: Good point!

dynobo:
> @sfalkner: We can try to solve this is this PR, or do that in a separate one, if we can tolerate a skipped tests :)

I already proposed a new function to solve this in #65. I'm going to rebase this branch here on top of that and try to apply the new function on BOHB to make it pass the test.

blackboxopt/optimizers/testing.py
LGro: I'm responsible for that pattern but don't see a reason to abandon it.

blackboxopt/optimizers/testing.py
```diff
@@ -148,6 +150,81 @@ def is_deterministic_with_fixed_seed(
...
+    optimizer_class: Union[
+        Type[SingleObjectiveOptimizer], Type[MultiObjectiveOptimizer]
+    ],
+    optimizer_kwargs: dict,
+) -> bool:
+    """Check if determinism isn't affected by the order of initially reported data.
+
+    Repeatedly initialize the optimizer with the same parameter space and a fixed seed.
+    Report a set of initial evaluations in randomized order as initial data. Start
+    optimizing and check if the generated configurations for all optimizers are equal.
```
LGro, line 163: Are we confident that reporting custom evaluations initially before optimization is the only relevant case we need to test for order invariance? In principle, the same situation can occur also at any other stage of optimization, consider e.g. parallel optimization runs where the runtime of the parallel evaluation jobs is not deterministic.

blackboxopt/optimizers/testing.py
```diff
@@ -148,6 +150,81 @@ def is_deterministic_with_fixed_seed(
...
+        optimizer_kwargs: Expected to contain additional arguments for initializing
+            the optimizer. (`search_space` and `objective(s)` are set automatically
+            by the test.)
+
+    Returns:
+        `True` if the test is passed.
+    """
+
+    space = ps.ParameterSpace()
+    space.add(ps.ContinuousParameter("p1", [0, 1]))
```
LGro, line 176: The bounds needs to be a tuple, not a list `(0, 1)` (I know we're not extremely good at sticking to this, but it's [true](https://github.com/boschresearch/parameterspace/blob/main/parameterspace/parameters/continuous.py#L25)).

blackboxopt/optimizers/testing.py
```diff
@@ -148,6 +150,81 @@ def is_deterministic_with_fixed_seed(
...
+        opt = _initialize_optimizer(
+            optimizer_class,
+            optimizer_kwargs,
+            objective=Objective("loss", False),
+            objectives=[Objective("loss", False)],
+            space=space,
+            seed=0,
+        )
+
+        # Report initial data in differenht order
```
LGro, line 197: typo: "differenht" -> "different"

LGro (approved):
Thanks @dynobo :relaxed:

blackboxopt/optimizers/testing.py
dynobo: >  In principle, the same situation can occur also at any other stage of optimization, consider e.g. parallel optimization runs where the runtime of the parallel evaluation jobs is not deterministic.

That's absolutely true, I'm very certain that the additional test is _not_ extensively covering all ordering related determinism issue, my motivation was only to extend coverage for the cases we observed so far.

Regarding the scenario of parallel optimization that you described, do you think there is a reasonable _reference_ test we could design? My understanding is, that the degree of determinism in such a case is very likely dependent on the type of optimizer (e.g. if the evaluations are processed staged/batched), is this correct?

I'm also not sure, to which degree we should _aim_ to be deterministic. That's something we might want to discuss in a bigger round, I guess.

Do you see any specific scenarios we _should_ cover here in _this_ PR?

LGro: If this covers the issues we've seen that's good enough for me to go ahead with this PR. Especially since you're spot on about the challenges to find a representative, general enough reference test that would apply and discover issues in more individual optimization behavior.

blackboxopt/optimizers/testing.py
dynobo: Done

blackboxopt/optimizers/testing.py
dynobo: Done

blackboxopt/optimizers/testing.py
dynobo: Done

blackboxopt/optimizers/testing.py
dynobo: Removed all of those.

PR #65, opened 2022-10-19
Add utility function to derministically sort list of evaluations

dynobo:
This doesn't introduce any functional changes yet, but prepares fixing determinism issues in #64.

The goal is to provide a function that's able to deterministically sort a `List[Evaluation]` by _only_ considering optimization relevant attributes, disregarding all others that might be subject to change/randomness (e.g. timestamps). 

Sorting the dictionary keys is the only option I found for being robust also against changes in the order of parameters, too. 

Performance depends a lot on the amount of parameters and the number of elements. In a quick test I tried different methods using a list of 1000 evaluations with 20 parameters.  The time is total seconds for sorting 1000 times (or the mean milliseconds per sort):

```
json sorted 
32.07862658100203

md5 of json sorted
29.56752566599971

md5 of pickle
2.9176704900019104

md5 of pickle of sorted <-- pull requested implementation
8.36911787000281

md5 of str
21.312109268001223

str
19.841941792001307
```

What do you think?

Also: I was unsure if we should include `optimizer_info` into the sorting. To me it seems like it's mostly used/set internally by the optimizer, but I'm not 100% certain.

### DIFF
--- a/blackboxopt/utils.py
+++ b/blackboxopt/utils.py
@@ -3,6 +3,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import hashlib
+import pickle
 from itertools import compress
 from typing import Dict, List, Optional, Sequence
 
@@ -80,3 +82,20 @@ def filter_pareto_efficient(
     pareto_efficient_mask = mask_pareto_efficient(losses)
 
     return list(compress(evaluations, pareto_efficient_mask))
+
+
+def sort_evaluations(evaluations: List[Evaluation]) -> List[Evaluation]:
+    return sorted(
+        evaluations,
+        key=lambda e: hashlib.md5(
+            pickle.dumps(
+                [
+                    sorted(e.configuration.items()),
+                    sorted(e.objectives.items()),
+                    sorted(e.settings.items()),
+                    sorted(e.constraints.items()) if e.constraints else {},
+                    sorted(e.context.items()) if e.context else {},
+                ]
+            )
+        ).hexdigest(),
+    )
--- a/tests/utils_test.py
+++ b/tests/utils_test.py
@@ -3,6 +3,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import random
+
 import numpy as np
 import pytest
 
@@ -11,6 +13,7 @@ from blackboxopt.utils import (
     filter_pareto_efficient,
     get_loss_vector,
     mask_pareto_efficient,
+    sort_evaluations,
 )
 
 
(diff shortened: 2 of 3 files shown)


attilareiss:
> Also: I was unsure if we should include `optimizer_info` into the sorting. To me it seems like it's mostly used/set internally by the optimizer, but I'm not 100% certain.

I would not include `optimizer_info` into the sorting. This field (potentially) includes data from optimization run dynamics, which might not be deterministic w.r.t. all other data (configuration, objectives) of the evaluations. For example, `evaluation_id` is a typical entry in `optimizer_info`, which only depends on the order in which evaluations are provided to the optimizer.

attilareiss (approved):
Looking at the test use cases, I believe all possible scenarios are nicely covered. Thanks @dynobo for the great addition!

tests/utils_test.py
```diff
@@ -89,3 +92,88 @@ def test_filter_pareto_efficient():
...
+            ),
+            Evaluation(
+                configuration={"i": 1}, objectives={"l": 0}, constraints={"c": 2}
+            ),
+            Evaluation(
+                configuration={"i": 1}, objectives={"l": 0}, constraints={"c": 3}
+            ),
+        ],
+    ),
+)
```
LGro, line 142: If we're going for saving space with single letter names, I'd encourage you to use the patter from settings and constraints also for objectives and parameter names. Specifically, `{"o": 1.0}` or `{"o1": 1.0, "o2": -2.0}` for objectives and `{"p": 3.0}` or `{"p1": 2.0, "p2": 1.0}` for configuration. That seems to reduce the necessity to visually jump back from the value to the corresponding kwarg.

blackboxopt/utils.py
```diff
@@ -80,3 +82,20 @@ def filter_pareto_efficient(
...
+                [
+                    sorted(e.configuration.items()),
+                    sorted(e.objectives.items()),
+                    sorted(e.settings.items()),
+                    sorted(e.constraints.items()) if e.constraints else {},
+                    sorted(e.context.items()) if e.context else {},
+                ]
+            )
+        ).hexdigest(),
+    )
```
LGro, line 101: Thanks a lot for evaluating different alternatives, I'd invite you to add a small ADR to document that reasoning.

LGro (approved):
Thanks @dynobo :relaxed:

tests/utils_test.py
dynobo: Jup, fine for me

dynobo: Done

blackboxopt/utils.py
dynobo: Done

PR #66, opened 2022-10-24
Feature/fix determinism in bohbs config sampler

dynobo:
While working on #64 and #65 I noticed, that BOHB's config sampler wasn't yet deterministic. This wasn't noticed so far, because the effect only becomes visible after multiple evaluations, and also seems to depend on the search space. (With our default mixed space for testing, there was no effect within 15 evaluations, while with a more simple 1D space, the proposed configurations started to diverge slightly after ~8 evaluations).

This PR proposes:

1. An additional reference test to capture this issue (which BOHB fails) 
2. A way to seed the BOHBs config sampler (to make BOHB pass the new test)

**Notes:** 
Unfortunately, `statsmodels` seems to still rely on global `numpy.random.seed()`, and the evaluation of the KDE estimators PDF depends on that. I tried to increase the thread safeness by (re-)seeding very close to the actual call. Also I cache the global state and revert it after our seeding to avoid undesired side effects (e.g. if a user depends on a certain global numpy seed). [This section still not looks really clean to me](https://github.com/boschresearch/blackboxopt/pull/66/files#diff-b0243323422e6ec3d5d243dff46be689c0a769ce03499e41ca7411ed8579fd17R311-R331), so if you have any better ideas, that'd be great. :-)

### DIFF
--- a/blackboxopt/optimizers/staged/bohb.py
+++ b/blackboxopt/optimizers/staged/bohb.py
@@ -55,8 +55,7 @@ def sample_around_values(
     Returns:
         Numerical representation of a configuration close to the provided datum.
     """
-    if rng is None:
-        rng = np.random.default_rng()
+    rng = np.random.default_rng(rng)
 
     vector = []
     for m, bw, t in zip(datum, bandwidths, vartypes):
@@ -151,8 +150,7 @@ def impute_conditional_data(
         Numerical representation where all NaNs have been replaced with observed values
         or prior samples.
     """
-    if rng is None:
-        rng = np.random.default_rng()
+    rng = np.random.default_rng(rng)
 
     return_array = np.empty_like(array)
 
@@ -322,7 +320,7 @@ class Sampler(StagedIterationConfigurationSampler):
                 cached_rng_state = None
                 if self.seed:
                     cached_rng_state = np.random.get_state()
-                    np.random.seed(self.seed + 1)
+                    np.random.seed(self.seed + i)
 
                 val = minimize_me(vector)
 
--- a/blackboxopt/optimizers/staged/bohb.py
+++ b/blackboxopt/optimizers/staged/bohb.py
@@ -33,6 +33,7 @@ def sample_around_values(
     vartypes: Union[list, np.ndarray],
     min_bandwidth: float,
     bw_factor: float,
+    rng: Optional[np.random.Generator] = None,
 ) -> Optional[np.ndarray]:
     """Sample numerical representation close to a given datum.
 
@@ -49,21 +50,27 @@ def sample_around_values(
             samples agree on a value in a dimension.
         bw_factor: To increase diversity, the bandwidth is actually multiplied by this
             factor before sampling.
+        rng: A random number generator to make the sampling reproducible.
 
     Returns:
         Numerical representation of a configuration close to the provided datum.
     """
+    if rng is None:
+        rng = np.random.default_rng()
+
     vector = []
     for m, bw, t in zip(datum, bandwidths, vartypes):
         bw = max(bw, min_bandwidth)
         if t == 0:
             bw = bw_factor * bw
             try:
-                v = sps.truncnorm.rvs(-m / bw, (1 - m) / bw, loc=m, scale=bw)
+                v = sps.truncnorm.rvs(
+                    -m / bw, (1 - m) / bw, loc=m, scale=bw, random_state=rng
+                )
             except Exception:
                 return None
         elif t > 0:
-            v = m if np.random.rand() < (1 - bw) else np.random.randint(t)
+            v = m if rng.random() < (1 - bw) else rng.integers(t)
         else:
             bw = min(0.9999, bw)  # bandwidth has to be less the one for this kernel!
             diffs = np.abs(np.arange(-t) - m)
@@ -71,7 +78,7 @@ def sample_around_values(
             idx = diffs == 0
             probs[idx] = (idx * (1 - bw))[idx]
             probs /= probs.sum()
-            v = np.random.choice(-t, p=probs)
+            v = rng.choice(-t, p=probs)
         vector.append(v)
     return np.array(vector)
 
@@ -125,7 +132,9 @@ def convert_from_statsmodels_kde_representation(
 
 
 def impute_conditional_data(
-    array: np.ndarray, vartypes: Union[list, np.ndarray]
+    array: np.ndarray,
+    vartypes: Union[list, np.ndarray],
+    rng: Optional[np.random.Generator] = None,
 ) -> np.ndarray:
     """Impute NaNs in numerical representation with observed values or prior samples.
 
@@ -137,11 +146,14 @@ def impute_conditional_data(
             values for inactive variables.
         vartypes: Encoding of the types of the variables: 0 mean continuous, >0 means
             categorical with as many different values, and <0 means ordinal with as many values.
-
+        rng: A random number generator to make the imputation reproducible.
     Returns:
         Numerical representation where all NaNs have been replaced with observed values
         or prior samples.
     """
+    if rng is None:
+        rng = np.random.default_rng()
+
     return_array = np.empty_like(array)
 
     for i in range(array.shape[0]):
@@ -154,7 +166,7 @@ def impute_conditional_data(
 
             if len(valid_indices) > 0:
                 # pick one of them at random and overwrite all NaN values
-                row_idx = np.random.choice(valid_indices)
+                row_idx = rng.choice(valid_indices)
                 datum[nan_indices] = array[row_idx, nan_indices]
 
             else:
@@ -162,11 +174,11 @@ def impute_conditional_data(
                 # but random value
                 t = vartypes[nan_idx]
                 if t == 0:
-                    datum[nan_idx] = np.random.rand()
+                    datum[nan_idx] = rng.random()
                 elif t > 0:
-                    datum[nan_idx] = np.random.randint(t)
+                    datum[nan_idx] = rng.integers(t)
                 elif t < 0:
-                    datum[nan_idx] = np.random.randint(-t)
+                    datum[nan_idx] = rng.integers(-t)
             nan_indices = np.argwhere(np.isnan(datum)).flatten()
         return_array[i, :] = datum
     return return_array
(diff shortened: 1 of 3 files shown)


blackboxopt/optimizers/staged/bohb.py
```diff
@@ -49,29 +50,35 @@ def sample_around_values(
             samples agree on a value in a dimension.
         bw_factor: To increase diversity, the bandwidth is actually multiplied by this
             factor before sampling.
+        rng: A random number generator to make the sampling reproducible.
 
     Returns:
         Numerical representation of a configuration close to the provided datum.
     """
+    if rng is None:
+        rng = np.random.default_rng()
```
sfalkner, line 59:
```suggestion
        rng = np.random.default_rng(rng)
```
IIRC, `np.random.default_rng` will return the global instance if the provided seed is None, and simply the input if a valid generator is passed. So I think the `if` can be dropped here.

blackboxopt/optimizers/staged/bohb.py
```diff
@@ -137,11 +146,14 @@ def impute_conditional_data(
...
         vartypes: Encoding of the types of the variables: 0 mean continuous, >0 means
             categorical with as many different values, and <0 means ordinal with as many values.
-
+        rng: A random number generator to make the imputation reproducible.
     Returns:
         Numerical representation where all NaNs have been replaced with observed values
         or prior samples.
     """
+    if rng is None:
+        rng = np.random.default_rng()
```
sfalkner, line 155:
```suggestion
    rng = np.random.default_rng(rng)
```
Same as above, I think.

blackboxopt/optimizers/staged/bohb.py
```diff
@@ -279,20 +295,40 @@ def sample_configuration(self) -> Tuple[dict, dict]:
...
+                # the same random numbers each time.
+                # We also reset the np.random's global state, in case the user relies
+                # on it in other parts of the code and to not hide other determinism
+                # issues.
+                # TODO: Check github issue if there was progress and the seeding can be
+                # removed: https://github.com/statsmodels/statsmodels/issues/306
+                cached_rng_state = None
+                if self.seed:
+                    cached_rng_state = np.random.get_state()
+                    np.random.seed(self.seed + 1)
```
sfalkner, line 325: Doesn't this line seed the rng for the statsmodel part always in the same way? I was expecting this seed to be changed somewhere to not get the same random numbers in each iteration. Hope I am missing something :)

blackboxopt/optimizers/testing.py
```diff
@@ -148,6 +150,55 @@ def is_deterministic_with_fixed_seed(
...
+            optimizer_kwargs,
+            objective=Objective("loss", False),
+            objectives=[Objective("loss", False)],
+            space=space,
+            seed=42,
+        )
+
+        for _ in range(15):
+            es = opt.generate_evaluation_specification()
+            evaluation = es.create_evaluation(objectives={"loss": 0.42})
```
sfalkner, line 193: Not sure if that is an issue, but having the same loss for all evaluations relies on the assumption that the sorting in BOHB is stable, which might be true or not. To fix this potential issue, simply using the iteration index as the loss, would make the sorting unique.

sfalkner (approved):
Thanks @dynobo for digging into this issue and fixing it :)

blackboxopt/optimizers/staged/bohb.py
dynobo: Ah, thanks! It has to be `self.seed + i` instead of `self.seed + 1`. Changed that temporarily for debugging and forgot to revert...

blackboxopt/optimizers/staged/bohb.py
dynobo: PS: Just FYI: I think I can't just "commit suggestion" on your "suggested changes", because those commits won't be signed. They are helpful nevertheless.

sfalkner: True, I forgot about that requirement. But if you don't mind, I will keep doing them in the cases where the suggestion is much simpler than trying to describe the change in the text.

dynobo: For sure, please keep doing it, wherever it seems helpful. As a requester, this is super useful as well! Just wanted to explain the reason for not committing suggestion right away... ;-)

blackboxopt/optimizers/testing.py
```diff
@@ -321,7 +372,8 @@ def handles_conditional_space(
 
 ALL_REFERENCE_TESTS = [
     optimize_single_parameter_sequentially_for_n_max_evaluations,
-    is_deterministic_with_fixed_seed,
+    is_deterministic_with_fixed_seed_and_mixed_space,
+    is_deterministic_with_fixed_seed_and_multiple_evaluations,
```
LGro, line 376: Is it the presence of multiple parameter types or in general a higher search space dimensionality that makes the difference here? If it's the dimensionality independent of the types, let's rename the reference test from `_and_mixed_space` to something along the lines of `_and_large_space` or `_and_4D_space` (or what ever the dimensionality is). Otherwise, seeing the two tests side by side, it might be misread as the parameter types making the difference. Whereas if that's indeed what makes the difference, everything is alright :)

blackboxopt/optimizers/testing.py
```diff
@@ -27,7 +27,7 @@ def _initialize_optimizer(
     seed=42,
 ) -> Optimizer:
     if space is None:
-        space = ps.ParameterSpace()
+        space = ps.ParameterSpace(seed=seed)
```
LGro, line 30: This should not be necessary, since the `Optimizer` base class [seeds its search space](https://github.com/boschresearch/blackboxopt/blob/main/blackboxopt/base.py#L148-L149) when a seed is given. Was there a specific reason (i.e. something failing) that motivated you to change this?

blackboxopt/optimizers/staged/bohb.py
```diff
@@ -125,7 +131,9 @@ def convert_from_statsmodels_kde_representation(
 
 
 def impute_conditional_data(
-    array: np.ndarray, vartypes: Union[list, np.ndarray]
+    array: np.ndarray,
+    vartypes: Union[list, np.ndarray],
+    rng: Optional[np.random.Generator] = None,
```
LGro, line 136: Is there a particular reason to be conservative here, and not allow integer seeds? Because the use in the function suggest that this could be the default numpy seed type with everything it allows.

blackboxopt/optimizers/staged/bohb.py
```diff
@@ -33,6 +33,7 @@ def sample_around_values(
     vartypes: Union[list, np.ndarray],
     min_bandwidth: float,
     bw_factor: float,
+    rng: Optional[np.random.Generator] = None,
```
LGro, line 36: same as [here](https://github.com/boschresearch/blackboxopt/pull/66/files#r1012570713)

LGro (approved):
Thanks @dynobo :rocket:

blackboxopt/optimizers/testing.py
dynobo: Actually, I don't know what makes the difference. I'll give it a try and rename accordingly.

blackboxopt/optimizers/testing.py
dynobo: Nope, just didn't know if/where the space was seeded...

blackboxopt/optimizers/staged/bohb.py
dynobo: > Is there a particular reason to be conservative here, and not allow integer seeds?

Yes, my motivation was, that re-seeding a function/logic that's called repeatedly could result in less random behavior. E.g. if `impute_conditional_data()` would be called with an integer seed instead of a Generator, I would have to ensure to input a _different_ seed at every call to ensure "real" random behavior across multiple calls of that function. I hope to avoid that be using a single RNG throughout the whole optimizer.

Does this make sense?

blackboxopt/optimizers/staged/bohb.py
dynobo: same as [here](https://github.com/boschresearch/blackboxopt/pull/66#discussion_r1012711411) ;-)

blackboxopt/optimizers/testing.py
LGro: Ok, then let's not seed it here in order to spot when optimizers don't call the super init properly.

blackboxopt/optimizers/staged/bohb.py
LGro: I definitely see how in the current use case all we (should) do is pass around a single RNG, but the function signature itself doesn't need to know that about its use, I'd say. My proposal would be to make the type hint as generic as it can be and still use the function in exactly the way we do with the current proposal.

dynobo: What do you see as benefit of having a generic type hint here?

I'd say a more precise type hint discourage inappropriate usage (e.g. passing an integer instead) a bit more.

blackboxopt/optimizers/testing.py
dynobo: Renamed

blackboxopt/optimizers/testing.py
dynobo: Done

blackboxopt/optimizers/staged/bohb.py
LGro: > What do you see as benefit of having a generic type hint here?

That it accurately represents all valid ways in which this function can be used.

> I'd say a more precise type hint discourage inappropriate usage (e.g. passing an integer instead) a bit more.

Which would be inappropriate in the current context it is used. I guess the difference in our view points is how much one considers reusability. Because if one only considers the only current use, then having the restrictive type hint can be appropriate. If one considers reusability, it's unnecessarily restrictive because the inappropriateness of passing an integer solely comes from the context and not the function itself. There can be call contexts where it's ok to use an int instead of an rng.
So we can either call scope creep or premature optimization, and there are valid arguments for both, none of which warrant delaying this PR for longer in my view :relaxed:

dynobo: Thanks for your explanation. I'll leave it as is for now, then and we can adjust the type hint once we reuse the methods in a situation where it really would be appropriate to call them with another type then the specified one.

PR #69, opened 2022-11-18
Extend deterministic ref test to compare five configurations.

aeivazi:
Signed-off-by: Anna Eivazi <anna.eivazi@de.bosch.com>

This PR proposes to increase number of configurations (requested sequentially) for checking optimizer determinism, which makes test more reliable and less dependent on number of initial sampling strategies or optimizer's specific heuristics for initial proposals.

### DIFF
--- a/blackboxopt/__init__.py
+++ b/blackboxopt/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "4.6.2"
+__version__ = "4.6.3"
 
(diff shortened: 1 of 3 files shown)


lreeb (approved):
Thanks @aeivazi !

dynobo (approved):
Thanks @aeivazi! :rocket:

PR #70, opened 2022-11-22
Add constraints

aeivazi:
Add constraints values while reporting evaluations to be able to test constraints aware optimizers. 

### DIFF
--- a/blackboxopt/__init__.py
+++ b/blackboxopt/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "4.6.3"
+__version__ = "4.6.4"
 
 from parameterspace import ParameterSpace
 
--- a/blackboxopt/optimizers/testing.py
+++ b/blackboxopt/optimizers/testing.py
@@ -85,7 +85,9 @@ def optimize_single_parameter_sequentially_for_n_max_evaluations(
             objectives={"loss": None, "score": None}
         )
     else:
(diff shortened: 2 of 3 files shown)


lreeb:
Sorry if that's not in the scope of this PR, but I noticed in [evaluation.py](https://github.com/boschresearch/blackboxopt/blob/cbfc85b059c724c73989ef9bd7358818e269674d/blackboxopt/evaluation.py#L66) constraints are described as 
>For each constraint name the float value indicates how much the constraint was satisfied, with negative values implying a violated and values indicating a satisfied constraint

But as far as I understood we would report the actual value of the constraint in the Evaluation and then the AquisitionFunction will take care of checking if constraints are met?

lreeb (approved):

aeivazi:
> Sorry if that's not in the scope of this PR, but I noticed in [evaluation.py](https://github.com/boschresearch/blackboxopt/blob/cbfc85b059c724c73989ef9bd7358818e269674d/blackboxopt/evaluation.py#L66) constraints are described as
> 
> > For each constraint name the float value indicates how much the constraint was satisfied, with negative values implying a violated and values indicating a satisfied constraint
> 
> But as far as I understood we would report the actual value of the constraint in the Evaluation and then the AquisitionFunction will take care of checking if constraints are met?

Yes, you are absolutely right. The text was motivated (some long time ago) by some other way to represent constraints. I am planning to discuss it internally soon and see which representation we will go for.

PR #71, opened 2022-11-30
add generic botorch base optimizer

LGro:
This PR brings a fairly generic BOTorch based optimizer to the table, which we can flexibly use with various models and acquisition functions. For a simple example, see the reference tests implementation.


pyproject.toml
```diff
@@ -24,6 +24,11 @@ statsmodels = {version = "^0.13.2", optional = true}
 dask = {version = "^2021.2.0", optional = true}
 distributed = {version = "^2021.2.0", optional = true}
 pandas = {version = "^1.2.4", optional = true}
+botorch = {version = "^0.7.1", optional = true}
+# version 1.13.0 has a bug where many unnecessary heavy dependencies are installed in
+# the base configuration of torch. Fixing it to <1.13 until the bug is resolved.
+# https://github.com/pytorch/pytorch/issues/88869
+torch = {version = "~1.12", optional = true}
```
LGro, line 31: The background on this is that since `1.13` all kinds of cuda dependencies come with installing `torch` and I'm wondering if the way how to get a cpu only version of torch changed to installing from https://download.pytorch.org/whl/cpu; Do you know anything about this @sfalkner?

blackboxopt/optimizers/botorch_base.py
```diff
@@ -0,0 +1,463 @@
...
+        self.af_opt_kwargs = init_af_opt_kwargs(af_optimizer_kwargs)
+
+    def _create_fantasy_model(self, model: Model) -> Model:
+        """Create model with the pending specifications and model based
+        outcomes added to the training data."""
+        if not self.pending_specifications:
+            # nothing to do when there are no pending specs
+            return model
+
+        pending_X = torch.tensor(
```
sfalkner, line 233: `pending_X` needs to be given `dtype=self.torch_dtype`, otherwise this will always end up being a `float64` tensor, as this is the numpy default.

blackboxopt/optimizers/botorch_base.py
```diff
@@ -0,0 +1,463 @@
...
+        X, Y = to_numerical(
+            _evals,
+            self.search_space,
+            self.objective,
+            self.batch_shape,
+            self.torch_dtype,
+        )
+
+        # fill in NaNs originating from inactive parameters (conditional spaces support)
+        X = impute_nans_with_constant(X)
```
aeivazi, line 367: Would it be ok to add already during this PR changes we've discussed last week:
```
# fill in NaNs originating from inactive parameters (conditional spaces support)
# botorch expect numerical representation of inputs to be within the unit hypercube
X = impute_nans_with_constant(X, c=0.0)
```

pyproject.toml
LGro: I'll change the dependency to install from that extra index URL.

aeivazi:
I think this PR is ready to be merged. I have no further comments. Thank you @LGro!

aeivazi (approved):

pyproject.toml
LGro: I don't seem to be able to find a way to take care of this without incredibly slowing down poetry update. When I specify a new source:
```toml
[[tool.poetry.source]]
name = "torch-cpu"
url = "https://download.pytorch.org/whl/cpu/"
```
and then specify for torch to be fetched from that source, it'll still check the torch repo for all other dependencies before going to pypi.org; this is mitigated to some extent by also specifying pypi explicitly as a source and setting all first level dependencies to be fetched from there, but for some (I guess higher level dependencies) we still run into the issue that poetry first checks the torch index, slowing down the update by ~10s per dependency.

My proposal would be to just go for the default gpu enabled torch, accepting the downside that setup takes longer and requires more disk space for the botorch extra. Thoughts, @aeivazi & @sfalkner?

LGro: Thanks for the suggestion, but without success I tried:
```toml
[[tool.poetry.source]]
name = "pypi_"
url = "https://pypi.python.org/simple"
default = true

[[tool.poetry.source]]
name = "torch-cpu"
url = "https://download.pytorch.org/whl/cpu/"
default = false
secondary = true

[tool.poetry.dependencies]
python = ">=3.8,<3.11"
parameterspace = "^0.7.2"
numpy = {version = "^1.22.0", optional = true, source = "pypi_"}
plotly = {version = "^5.10.0", optional = true, source = "pypi_"}
scipy = {version = "^1.6.0", optional = true, source = "pypi_"}
statsmodels = {version = "^0.13.2", optional = true, source = "pypi_"}
dask = {version = "^2021.2.0", optional = true, source = "pypi_"}
distributed = {version = "^2021.2.0", optional = true, source = "pypi_"}
pandas = {version = "^1.2.4", optional = true, source = "pypi_"}
botorch = {version = "^0.7.1", optional = true, source = "pypi_"}
torch = {version = "^1.11", optional = true, source = "torch-cpu"}


[tool.poetry.dev-dependencies]
black = {version = "^22.3.0", source = "pypi_"}
isort = {version = "^5.10.1", source = "pypi_"}
pytest = {version = "^7.2.0", source = "pypi_"}
mypy = {version = "^0.950", source = "pypi_"}
pytest-timeout = {version = "^2.1.0", source = "pypi_"}
pytest-cov = {version = "^3.0.0", source = "pypi_"}
mkdocs = {version = "^1.3.0", source = "pypi_"}
mkdocstrings = {version="^0.18.1", extras=["python-legacy"], source = "pypi_"}
mkdocs-material = {version = "^8.2.11", source = "pypi_"}
mkdocs-gen-files = {version = "^0.3.4", source = "pypi_"}
mkdocs-awesome-pages-plugin = {version = "^2.7.0", source = "pypi_"}
mike = {version = "^1.1.2", source = "pypi_"}
pylint = {version = "^2.13.8", source = "pypi_"}
pydocstyle = {version = "^6.1.1", source = "pypi_"}
pre-commit = {version = "^2.18.1", source = "pypi_"}
toml = {version = "^0.10.2", source = "pypi_"}
types-toml = {version = "^0.10.7", source = "pypi_"}
kaleido = {version = "0.2.1", source = "pypi_"} # pin hard, because 0.2.1post1 fails to install
```

sfalkner: Is the package name `torch-cpu` or simply `torch`? The pip command from the official pytorch docs reads
```pip3 install torch torchvision torchaudio --extra-index-url https://download.pytorch.org/whl/cpu```
which makes me wonder if the package name should simply be `torch`.

**Review by sfalkner — APPROVED** [review]
Also no further comments from my side. Thanks a lot @LGro for this effort!

File: pyproject.toml (hunk 1 above)
Comment (LGro) [reply pyproject.toml]: The package name is `torch`, the source index name we gave in our config was `torch-cpu` but this was fully in our scope and has nothing to do with anything the torch maintainers offer.

Comment (LGro) [reply pyproject.toml]:
```
torch = {version = "^1.11", optional = true, source = "torch-cpu"}
```
is where the dependency `torch` is defined, and `torch-cpu` is just a local name to reference to the extra index `https://download.pytorch.org/whl/cpu/`.

sfalkner: Thanks for clarifying and sorry for my confusion.

PR #73, opened 2022-12-07
add py3.11 to compatible versions and CI/CD test

LGro:
Currently, `dask.distributed` still seems to struggle with Py3.11 or need special treatment. Thus, opening the support for Python 3.11 as a draft for now.

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -41,7 +41,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04]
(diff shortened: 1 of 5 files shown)


LGro:
Now the issue seems to be about dask compatibility of the interface during test, so I think we're beyond the original problem of missing Py3.11 compatibility.

dynobo:
I'll take a look at it now.

LGro:
Thanks for jumping in @dynobo :relaxed:

dynobo (approved):

PR #75, opened 2022-12-09
Add constraint names support for `to_numerical` function

aeivazi:

### DIFF
--- a/blackboxopt/__init__.py
+++ b/blackboxopt/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "4.7.2"
+__version__ = "4.8.0"
 
 from parameterspace import ParameterSpace
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "blackboxopt"
-version = "4.7.2"
+version = "4.8.0"
 description = "A common interface for blackbox optimization algorithms along with useful helpers like parallel optimization loops, analysis and visualization scripts."
 readme = "README.md"
 repository = "https://github.com/boschresearch/blackboxopt"


blackboxopt/optimizers/botorch_base.py
```diff
@@ -77,30 +78,33 @@ def to_numerical(
...
     Returns:
         - X: Numerical representation of the configurations
-        - Y: Numerical representation of the objective values
+        - Y: Numerical representation of the objective values and optionally constraints
 
     Raises:
         ValueError: If one of configurations is not valid w.r.t. search space.
         ValueError: If one of configurations includes parameters that are not part of
             the search space.
+        ConstraintError: If one of the constraint names is not defined in evaluations.
```
aeivazi, line 107: Here is an interesting discussion, is `ConstraintError` raising fit here? or should it be a pure `ValueError` instead. 
Constraint Error docs says:
```
Raised on incomplete or missing constraints.
```
Which is not exactly what is happening here. We are raising an error if user requested a constraint that is not defined in input evaluations. What do you think?

dynobo: You are right, `ConstraintsError` doesn't seem correct here.

To me it feels like preserving the `KeyError` would actually fit quite well? If we want to preserve the custom error message, we could go with e.g. L152-155:
```python
        except KeyError as e:
            raise KeyError(
                f"Constraint name {e} is not defined in input evaluations."
            ) from e
```
PS: `raise ... from ...` might be a good idea anyway, because it retains the information provided by the initial exception

lreeb: What do you mean by "the user requests a constraint"? I thought `to_numerical` gets called when evaluations are reported by the user. The section below then tries to extract all constraints that the optimizer was initialized with (`constraint_names`) from each reported evaluation's constraints `e.constraints`:
```
[[e.constraints[c] for c in constraint_names] for e in evaluations]
```
If a constraint is missing in the evaluation that was defined for the optimizer, the error is thrown. So to me it seems like "Raised on incomplete or missing constraints." describes this pretty well? 
(Actually, it should probably also catch "AttributeError" which is thrown when the evaluation does not containt constraints at all...)

aeivazi: I agree with the way you describe @lreeb from optimizers point of view. But if i think about this function independently i actually disagree. 
In a nutshell, the function is filtering evaluations with (provided by user) constraint names. So, if a constraint name is not found, it is not a problem of an evaluation definition but it is a problem of user, that s/he did give a valid name. 

Great hint about `AttributeError`! I will test. 

@dynobo i tend to disagree about re-raise a `KeyError`. I would propose to raise `ValueError` instead. A dictionary structure is quite internal to evaluation object and i think it would be confusing for user to try to figure our from which dictionary it is coming from.

lreeb: Ah ok, now I understand your point of view! Since it's a public function in principle it can be called with any `constraint_names` that might not fit the evaluations. However I still think the main reason this error will happen is when a user reports evaluations with missing constraints, so I would like it if the error message would provide a clear hint for that. I have no strong opinion on whether it should be a `ConstraintsError` or `ValueError` though. Could you also just change the documentation of `ConstraintsError` or do we want to keep that unchanged?

lreeb (approved):
So it wasn't AttributeError but TypeError! Thanks for adding. Looks good to me now :+1:

PR #76, opened 2022-12-09
Ensure that bounds are of the same type as an optimizer data

aeivazi:

While investigating why switching to different default type within optimizer, i've found out, that for some models (e.g. `ModelList`) there is an error in type mismatch during AF optimization. Fixing bounds type to the optimizer type fixes it.


sfalkner (approved):
Thanks @aeivazi for spotting this! Hopefully that was the last place where we missed the typing :)

PR #78, opened 2022-12-16
Remove global seeding and add test fixture for seed

lreeb:
With our recent issues with flaky tests and global seeding, I propose to use a pytest fixture to create a random (but fixed) seed for each of the reference tests.
Benefits:
* Removes all global seeding
* Avoids runing tests with always the same fixed seeds -> will spot problems earlier
* If a test fails, the seed will show up in the log and the test can be reproduced easily

What do you think?

Signed-off-by: Reeb Lucia <Lucia.Reeb@de.bosch.com>

### DIFF
--- a/tests/optimizers/conftest.py
+++ b/tests/optimizers/conftest.py
@@ -2,6 +2,7 @@
 # see the NOTICE file and/or the repository https://github.com/boschresearch/blackboxopt
 #
 # SPDX-License-Identifier: Apache-2.0
+import random
 
 import parameterspace as ps
 import pytest
@@ -11,6 +12,11 @@ from blackboxopt import Evaluation
 constraint_name_1 = "constraint1"
 constraint_name_2 = "constraint2"
 objective_name = "objective"
+
+
+@pytest.fixture
+def seed():
+    return random.getrandbits(32)
 
 
 @pytest.fixture
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Custom
 .vscode/
+.idea/
 poetry.toml
 dask-worker-space/
 
--- a/architecture_decision_records/random-seed-for-reference-tests.md
+++ b/architecture_decision_records/random-seed-for-reference-tests.md
@@ -1,0 +1,17 @@
+### Context
+While our suite of unit tests should run in a stable and deterministic manner,
+running the tests under stochastic conditions allows us to detect instabilities
+that would otherwise remain undetected. Especially for the set of reference tests,
+we would like to ascertain that they run stable even with random seeds.
+In addition, we would like to mitigate the issues that arise
+from some of the optimizers setting a global (torch) seed, which is affecting
+subsequent tests.
+
+### Decision
+Instead of using a constant (fixed) seed, we will seed all reference tests with a
+random seed from a test fixture. That ensures that all tests will be fully reproducible,
+since the seed will be displayed in the test output.
+
+### Consequences
+We accept that our test suite will not be fully deterministic any more, and that it
+is possible that instabilities will surface in the existing tests.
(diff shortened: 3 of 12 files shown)


tests/optimizers/conftest.py
```diff
@@ -13,6 +14,11 @@
 objective_name = "objective"
 
 
+@pytest.fixture
+def seed():
+    return random.getrandbits(32)
```
dynobo, line 19: I'm curious: What was your reason for using `random.getrandbits(32)` over `random.randint()`? (I didn't know `getrandbits` before and had to look it up)

dynobo (approved):
Thanks Lucia, removing the global seed makes so much sense!
I would have chosen even a fixed seed for the unit tests, but I'm also super fine with trying a random one. If this should turn out to be too fragile and blocks PRs too often, we now easily can switch to a fixed seed :-)

dynobo:
If you have some time, you could also do this for [blackboxopt](https://github.com/boschresearch/blackboxopt/blob/main/blackboxopt/optimizers/testing.py). Otherwise just leave me a note, I'd also happily do it.

tests/optimizers/conftest.py
sfalkner: I guess the problem is that python integers are of arbitrary size and `random.randint()` needs to arguments (min and max) value. This way, we get an unsigned int with the usual 32bit size, which should work for all packages we use, i.e. numpy and torch.

sfalkner (approved):
Thanks @lreeb for this addition. Looks like we should use this in other places, too :)

lreeb:
> Thanks Lucia, removing the global seed makes so much sense! I would have chosen even a fixed seed for the unit tests, but I'm also super fine with trying a random one. If this should turn out to be too fragile and blocks PRs too often, we now easily can switch to a fixed seed :-)

I think this is somewhat of a fundamental decision that we have to make - using a fixed seed would make all of our tests deterministic, thus we wouldn't have to check for flaky tests any more, and have more stability. Using a random seed will probably take some more effort to make all tests stable - but will possibly detect more errors. I don't have a strong opinion on which way to go, just didn't like the in-between status right now!

LGro:
Given the concern that we need to test stochastic things, it seems like a very good idea to avoid fixed seeds, but to use random seeds so that we get the stochasticity as well as reproducibility. Regarding flaky tests, it's a good idea for us to retain a close look at how flaky specific optimizers are because we might need to do something about it at a certain threshold; and with random seeds that threshold becomes explicit instead of hiding behind a fixed seed that works.
Might it be worth recording our reasoning behind the seeding strategy in an ADR?

dynobo:
> Given the concern that we need to test stochastic things, it seems like a very good idea to avoid fixed seeds, but to use random seeds so that we get the stochasticity as well as reproducibility. Regarding flaky tests, it's a good idea for us to retain a close look at how flaky specific optimizers are because we might need to do something about it at a certain threshold; 

From my experience, deterministic unit/integration tests are very important for a fluent development workflow and to identify problems fast. I'd rather _not_ want to ask myself "Failed the test because of my change or because it is flaky?".

I agree, that testing stochasticity is important as well, but I see that as a different purpose and would rather see such potential flaky tests it in an extra, separated test-set. Especially, if we see, how bad we currently are in handling failures in those (see Teams Bots channel).

lreeb:
Good point about adding an ADR! 
@dynobo While I fully agree on the point about deterministic tests, that would also imply that we would need to seed almost every single test, since most of them use stochasticity in some way.
I'd vote on trying with random seeds for now, and if it turns out that this is not feasible or too much overhead we might revert to a fixed seed and split up the test suite in a deterministic and flaky part.

PR #81, opened 2023-02-10
fix model update in gpbo, add nan filtering

LGro:
Despite conditioning the model on data when reporting for the GPBO optimizer, we never stored the updated model. Thus, we also missed the lack of support for handling `None` objectives, which is also now there.


sfalkner (approved):
Thanks @LGro for fixing this!

PR #82, opened 2023-02-10
get __version__ from importlib instead of hard coding

LGro:
Because I forgot to manually update the `__version__` in accordance with `pyproject.toml` every single time for the last couple PRs, let's try this instead :relaxed:

### DIFF
--- a/blackboxopt/__init__.py
+++ b/blackboxopt/__init__.py
@@ -1,4 +1,6 @@
-__version__ = "4.8.4"
+import importlib
+
+__version__ = importlib.metadata.version("blackboxopt")
 
 from parameterspace import ParameterSpace
 


dynobo:
I feel your pain and am open try other things. But just to make sure, we have everything on the radar:

1. Why didn't the [test](https://github.com/boschresearch/blackboxopt/blob/main/tests/package_test.py#L8) capture this before the commit? 
2. If I remember correctly, the main reason for not using `importlib` was, that it uses the version of the _installed_ package, which might be different than the version you are currently working on (e.g. I think if you switch to a branch with a different version without running a new `poetry install`, it will report the wrong version). I don't see a problem regarding that in `blackboxopt`, but I think this is different for other repos we have, and I see a certain benefit of coherence among them. Still it might be worth to give up that coherence, if the benefit is high enough.
3. An alternative, which I usually use in other projects, might be to introduce a tool like `bumpversion` or maybe slightly more modern [`tbump`](https://github.com/your-tools/tbump) to update the version string in multiple places.

Does it still feels worth a try? Then let's do it.

dynobo (approved):

LGro:
> 1. Why didn't the [test](https://github.com/boschresearch/blackboxopt/blob/main/tests/package_test.py#L8) capture this before the commit?

Oh, it did, but only during CI/CD because I didn't run the whole test suite locally, just the parts that I knew I touched.

>     2. If I remember correctly, the main reason for not using `importlib` was, that it uses the version of the _installed_ package, which might be different than the version you are currently working on (e.g. I think if you switch to a branch with a different version without running a new `poetry install`, it will report the wrong version). I don't see a problem regarding that in `blackboxopt`, but I think this is different for other repos we have, and I see a certain benefit of coherence among them. Still it might be worth to give up that coherence, if the benefit is high enough.

Do you have a scenario in mind where this would go wrong and how? Because it seems to me that to access `__version__` you'd always have to have the package installed, and hence would need to have the package version accessible. One scenario that I'm suspecting might not be straight forward is when `blackboxopt` is installed as a dependency in edit mode, but I have difficulties judging the risk in that scenario.

>     3. An alternative, which I usually use in other projects, might be to introduce a tool like `bumpversion` or maybe slightly more modern [`tbump`](https://github.com/your-tools/tbump) to update the version string in multiple places.

This seems like the more common pattern, which I'd also be happy to try instead. Let's see what you have in mind for point 2 above, and if we get to a better idea about the risk we can make a more informed choice.

lreeb:
As someone who's also stumbled across this every time I made a PR - 

> Why didn't the [test](https://github.com/boschresearch/blackboxopt/blob/main/tests/package_test.py#L8) capture this before the commit?

I don't usually run all tests when doing a small change like bumping the versions.

> An alternative, which I usually use in other projects, might be to introduce a tool like bumpversion or maybe slightly more modern [tbump](https://github.com/your-tools/tbump) to update the version string in multiple places.

That might work, another alternative could be adding a custom pre-commit hook that checks if the version has been updated correctly.

LGro:
Going to introduce tbump instead, closing.

PR #83, opened 2023-02-10
Feature/choose af optimization strategy

LGro:
After introducing discrete space detection to switch to random choice based acquisition function optimization in #79, this PR now adds a detection mechanism to override this automated af optimizer choice in case any of the af optimizer specific required kwargs are set. I disregarded the alternative of exposing a more explicit additional init argument to choose between e.g. `af_optimizer="lbfgs"` and `af_optimizer="random"` to avoid adding one more init argument and thereby raise the conflict potential with the `af_optimizer_kwargs` init argument. I'm happy to discuss this trade-off.
NOTE: The first commit moved a couple of general utilities out of the optimizer module, the second commit contains the main changes for this PR.

### DIFF
--- a/blackboxopt/optimizers/botorch_base.py
+++ b/blackboxopt/optimizers/botorch_base.py
@@ -6,23 +6,19 @@
 import functools
 import logging
 import warnings
-from typing import Callable, Dict, Iterable, List, Optional, Tuple, Union
+from typing import Callable, Dict, Iterable, Optional, Tuple, Union
 
 from gpytorch.models import ExactGP
 
-from blackboxopt import (
-    ConstraintsError,
-    Evaluation,
-    EvaluationSpecification,
+from blackboxopt.base import (
     Objective,
     OptimizerNotReady,
-    sort_evaluations,
-)
-from blackboxopt.base import (
     SingleObjectiveOptimizer,
     call_functions_with_evaluations_and_collect_errors,
     validate_objectives,
 )
+from blackboxopt.evaluation import Evaluation, EvaluationSpecification
+from blackboxopt.utils import sort_evaluations
 
 try:
     import numpy as np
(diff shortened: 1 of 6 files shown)


blackboxopt/optimizers/botorch_base.py
```diff
@@ -51,13 +51,14 @@ def _acquisition_function_optimizer_factory(
...
-    not overridden by `af_opt_kwargs`.
+    not overridden by `af_opt_kwargs`. If any of the af optimizer specific required
+    kwargs are set, this overrides the automatic discrete space detection.
 
     Args:
         search_space: Search space used for optimization.
         af_opt_kwargs: Acquisition function optimizer configuration, e.g. containing
-            values for `n_samples` for discrete optimization, and `num_restarts`,
-            `raw_samples` for the continuous optimization case.
+            values for `num_choices_discrete` for discrete optimization, and
```
sfalkner, line 60: I find the name `num_choices_discrete` slightly confusing. It could be mistaken for the number of discretisation values in, e.g. grid search. More appropriate names could be `num_samples_discrete`, `num_discrete_samples`, `raw_samples_discrete` or similar.

sfalkner (approved):
Thanks @LGro for implementing this!

dynobo (approved):
Looks good!

blackboxopt/optimizers/botorch_base.py
LGro: I feel you and started with samples but wanted to be close to the argument names in the botorch function, which calls it choices. What do you think?

sfalkner: I can see why it's called choices in BOTorch, because it is already the list of configs from which to choose. We can go with `num_choices_discrete`, it just sounded odd, but if it's close to BOTorch naming, I guess it makes sense.

LGro: We could also do `num_random_choices` to get the discrete out if that's the problematic one, while still keeping the link to the botorch choices argument.

sfalkner: I like `num_random_choices` the best, so I vote for that :)

PR #85, opened 2023-02-15
Reduce precision to 1e-3 while comparing configurations for the test …

aeivazi:
…is_deterministic_when_reporting_shuffled_evaluations. For some algorithms and for some seeds, configurations are only equal only with a lower precision.


LGro (approved):
For parameter bounds `(0, 1)` calling it a match when the difference between parameter values is below 0.0015 sounds fair to me. Thanks @aeivazi :relaxed:

PR #86, opened 2023-02-23
Feature/optimize over entire discrete search spaces

ptighin:
@LGro here is a proposal of how optimizing over entire discrete spaces could work. Looking forward to your feedback :)

### DIFF
--- a/blackboxopt/optimizers/botorch_base.py
+++ b/blackboxopt/optimizers/botorch_base.py
@@ -9,6 +9,7 @@ import warnings
 from typing import Callable, Dict, Iterable, Optional, Tuple, Union
 
 from gpytorch.models import ExactGP
+from parameterspace import ParameterSpace
 
 from blackboxopt.base import (
     Objective,
@@ -44,6 +45,21 @@ except ImportError as e:
     ) from e
 
 
+def _get_numerical_points_from_discrete_space(space: ParameterSpace) -> np.ndarray:
+    """Retrieve all points from a discrete space in the numerical representation"""
+    points_along_dimensions = []
+    for parameter_name in space.get_parameter_names():
+        num_values = space.get_parameter_by_name(parameter_name)[
+            "parameter"
+        ].num_values  # type:ignore
+        points_along_dimensions.append(
+            np.linspace(1 / (2 * num_values), 1 - 1 / (2 * num_values), num_values)
+        )
+    points = np.meshgrid(*points_along_dimensions)
+    points = [p.reshape((p.size, 1)) for p in points]
+    return np.concatenate(points, axis=-1)
+
+
 def _acquisition_function_optimizer_factory(
     search_space: ps.ParameterSpace,
     af_opt_kwargs: Optional[dict],
@@ -75,7 +91,7 @@ def _acquisition_function_optimizer_factory(
         "num_restarts" in kwargs
         or "raw_samples" in kwargs
         or space_has_continuous_parameters
-    ):
+    ):  # continuous AF optimization
         return functools.partial(
             optimize_acqf,
             q=1,
@@ -86,12 +102,20 @@ def _acquisition_function_optimizer_factory(
             **kwargs,
         )
 
-    choices = torch.Tensor(
-        [
-            search_space.to_numerical(search_space.sample())
-            for _ in range(kwargs.pop("num_random_choices", 5_000))
-        ]
-    ).to(dtype=torch_dtype)
+    if "num_random_choices" not in kwargs and not space_has_continuous_parameters:
+        # Optimize over the entire discrete search space, if the number of random
+        # choices is not specified
+        choices = torch.from_numpy(
+            _get_numerical_points_from_discrete_space(search_space)
+        ).to(torch_dtype)
+    else:
+        # Optimize over the desired number of samples from the discrete search space
+        choices = torch.Tensor(
+            [
+                search_space.to_numerical(search_space.sample())
+                for _ in range(kwargs.pop("num_random_choices", 5_000))
+            ]
+        ).to(dtype=torch_dtype)
     return functools.partial(optimize_acqf_discrete, q=1, choices=choices, **kwargs)
 
 
--- a/blackboxopt/optimizers/botorch_base.py
+++ b/blackboxopt/optimizers/botorch_base.py
@@ -49,12 +49,24 @@ def _get_numerical_points_from_discrete_space(space: ParameterSpace) -> np.ndarr
     """Retrieve all points from a discrete space in the numerical representation"""
     points_along_dimensions = []
     for parameter_name in space.get_parameter_names():
-        num_values = space.get_parameter_by_name(parameter_name)[
+        parameter = space.get_parameter_by_name(parameter_name)[
             "parameter"
-        ].num_values  # type:ignore
-        points_along_dimensions.append(
-            np.linspace(1 / (2 * num_values), 1 - 1 / (2 * num_values), num_values)
-        )
+        ]  # type:ignore
+        if isinstance(parameter, ps.IntegerParameter):
+            bounds = (parameter.bounds[0], parameter.bounds[1] + 1)
+            points_along_dimensions.append(
+                [parameter.val2num(v) for v in range(*bounds)]
+            )
+        elif isinstance(parameter, ps.OrdinalParameter) or isinstance(
+            parameter, ps.CategoricalParameter
+        ):
+            points_along_dimensions.append(
+                [parameter.val2num(v) for v in parameter.values]
+            )
+        else:
+            raise ValueError(
+                f"Only discrete parameters are allowed but got {parameter}"
+            )
     points = np.meshgrid(*points_along_dimensions)
     points = [p.reshape((p.size, 1)) for p in points]
     return np.concatenate(points, axis=-1)
@@ -68,7 +80,10 @@ def _acquisition_function_optimizer_factory(
     """Prepare either BoTorch's `optimize_acqf_discrete` or `optimize_acqf` depending
     on whether the search space is fully discrete or not and set required defaults if
     not overridden by `af_opt_kwargs`. If any of the af optimizer specific required
-    kwargs are set, this overrides the automatic discrete space detection.
+    kwargs are set, this overrides the automatic discrete space detection. In case an
+    exclusively discrete space is detected and `num_random_choices` is not specified
+    in `af_opt_kwargs`, the discrete acquisition function optimizer is using all
+    possible combinations in the discrete space.
 
     Args:
         search_space: Search space used for optimization.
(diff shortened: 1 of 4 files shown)


blackboxopt/optimizers/botorch_base.py
```diff
@@ -44,6 +45,21 @@
...
+        ].num_values  # type:ignore
+        points_along_dimensions.append(
+            np.linspace(1 / (2 * num_values), 1 - 1 / (2 * num_values), num_values)
+        )
+    points = np.meshgrid(*points_along_dimensions)
+    points = [p.reshape((p.size, 1)) for p in points]
+    return np.concatenate(points, axis=-1)
+
+
 def _acquisition_function_optimizer_factory(
```
LGro, line 63: Let's add another sentence to the function docstring explaining that e.g. "In case an exclusively discrete space is detected and `num_random_choices` is not specified, the discrete acquisition function optimizer is using all possible combinations in the discrete space."

tests/optimizers/botorch_base_test.py
```diff
@@ -118,14 +127,6 @@ def test_acquisition_function_optimizer_factory_force_continuous():
...
     assert af_opt.func == optimize_acqf  # pylint: disable=no-member
 
-    af_opt = _acquisition_function_optimizer_factory(
-        discrete_space,
-        af_opt_kwargs={"raw_samples": 5_000},
-        torch_dtype=torch.float64,
-    )
-
-    assert af_opt.func == optimize_acqf  # pylint: disable=no-member
-
```
LGro, line 128: Why do we need to delete this part of the test? It seems to me like this is still relevant and we would benefit from keeping it around.

blackboxopt/optimizers/botorch_base.py
```diff
@@ -75,7 +91,7 @@ def _acquisition_function_optimizer_factory(
         "num_restarts" in kwargs
         or "raw_samples" in kwargs
         or space_has_continuous_parameters
-    ):
+    ):  # continuous AF optimization
         return functools.partial(
```
LGro, line 95: We usually discourage inline comments, following the [PEP recommendation](https://peps.python.org/pep-0008/#inline-comments) to use inline comments sparingly, and rather comment on a new line before the affected line.
In this case that would look like
```python
        or space_has_continuous_parameters
    ):
        # continuous AF optimization
        return functools.partial(
```
or
```python
    # continuous AF optimization
    if "num_random_choices" not in kwargs and (
        "num_restarts" in kwargs
        or "raw_samples" in kwargs
        or space_has_continuous_parameters
    ):
```

blackboxopt/optimizers/botorch_base.py
```diff
@@ -44,6 +45,21 @@
...
+def _get_numerical_points_from_discrete_space(space: ParameterSpace) -> np.ndarray:
+    """Retrieve all points from a discrete space in the numerical representation"""
+    points_along_dimensions = []
+    for parameter_name in space.get_parameter_names():
+        num_values = space.get_parameter_by_name(parameter_name)[
+            "parameter"
+        ].num_values  # type:ignore
+        points_along_dimensions.append(
+            np.linspace(1 / (2 * num_values), 1 - 1 / (2 * num_values), num_values)
+        )
```
LGro, line 57: This line seems to rely on the internal implementation details of parameterspace and duplicate instead of reuse, also limiting the flexibility to specify custom transformations in the given parameterspace. I'm wondering if such a more explicit alternative that relies on the public parameterspace interface is better suited:
```python
parameter = space.get_parameter_by_name(parameter_name)["parameter"]
if isinstance(parameter, ps.IntegerParameter):
    points_along_dimensions.apend([parameter.val2num(v) for v in range(*parameter.bounds)])
elif isinstance(parameter, ps.OrdinalParameter) or isinstance(parameter, ps.CategoricalParameter):
    points_along_dimensions.append([parameter.val2num(v) for v in parameter.values])
else:
    raise ValueError(f"Only discrete parameters are allowed but got {parameter}")
```

tests/optimizers/botorch_base_test.py
```diff
@@ -155,3 +156,28 @@ def test_find_optimum_in_1d_discrete_space(seed):
     assert (
         sum(l == 0 for l in losses) > 5
     ), "After figuring out the best of the three points, it should only propose that."
+
+
+def test_get_numerical_points_from_discrete_space():
+    p0l, p0h = -5, 10
```
LGro, line 162: `p0l` and `p0h` seem to be unnecessarily difficult to understand at first glance, since it seems to me we wouldn't loose anything by instead going with `p0_bounds = (-5, 10)`, what do you think?

LGro: There might be something even better, let's ignore this proposal.

tests/optimizers/botorch_base_test.py
```diff
@@ -155,3 +156,28 @@ def test_find_optimum_in_1d_discrete_space(seed):
...
+    p1 = ("small", "medium", "large")
+    p2 = ("woof", "miaow", "moo")
+    discrete_space = ps.ParameterSpace()
+    discrete_space.add(ps.IntegerParameter("integ", (p0l, p0h)))
+    discrete_space.add(ps.OrdinalParameter("ordin", p1))
+    discrete_space.add(ps.CategoricalParameter("categ", p2))
```
LGro, line 168: The variables for the bounds and values/choices seem to be very basically named and below we hard code their ranges and lengths. To improve this situation overall, how about we go for
```python
p_integ = ps.IntegerParameter("integ", (p0l, p0h))
discrete_space.add(p_integ)
p_ordin = ps.OrdinalParameter("ordin", p1)
discrete_space.add(p_ordin)
p_categ = ps.CategoricalParameter("categ", p2)
discrete_space.add(p_categ)
```
which would enable us to below do the more explicit and maintenance friendly
```python
assert points.shape[0] == p_integ.num_values * p_ordin.num_values * p_categ.num_values
```

tests/optimizers/botorch_base_test.py
```diff
@@ -155,3 +156,28 @@ def test_find_optimum_in_1d_discrete_space(seed):
...
+        assert (
+            (
+                points
+                == discrete_space.to_numerical(
+                    dict(integ=integ, ordin=ordin, categ=categ)
+                )
+            )
+            .all(axis=1)
+            .any()
+        )
```
LGro, line 183: I understand the need for a for loop here, but wonder if we can provide a more helpful assertion error message in case this fails. From developing this functionality, do you have an idea what would help or whether the default assertion error is already helpful enough?

tests/optimizers/botorch_base_test.py
```diff
@@ -155,3 +156,28 @@ def test_find_optimum_in_1d_discrete_space(seed):
...
+    p1 = ("small", "medium", "large")
+    p2 = ("woof", "miaow", "moo")
+    discrete_space = ps.ParameterSpace()
+    discrete_space.add(ps.IntegerParameter("integ", (p0l, p0h)))
+    discrete_space.add(ps.OrdinalParameter("ordin", p1))
+    discrete_space.add(ps.CategoricalParameter("categ", p2))
+
+    points = _get_numerical_points_from_discrete_space(discrete_space)
+    assert points.shape[0] == 16 * 3 * 3
+    assert points.shape[-1] == 3
```
LGro, line 172: I appreciate that the combination of the overall shape test and the `itertools.product` test combined ensure that nothing is missing but also no duplicates are contained, cool!

LGro, line 172: Please use `assert points.shape[-1] == len(discrete_space)` instead, because it helps folks understand where the expected value comes from.

LGro:
Welcome @ptighin, and thanks a lot for your first contribution to `blackboxopt` :relaxed:

tests/optimizers/botorch_base_test.py
```diff
@@ -155,3 +156,28 @@ def test_find_optimum_in_1d_discrete_space(seed):
...
+    for integ, ordin, categ in itertools.product(range(p0l, p0h), p1, p2):
```
LGro, line 173: Note, that `range(p0l, p0h)` is excluding the upper bound, while parameterspace uses inclusive bounds. See [here](https://github.com/boschresearch/blackboxopt/pull/86/files#r1116652802) for an alternative proposal.

blackboxopt/optimizers/botorch_base.py
ptighin: Excellent point. I expected that part of the code could be reused but was not sufficiently familiar with the internal workings of `parameterspace`. I implemented your suggestion with a small difference since `range(*parameter.bounds)` needs to include the last element.
```python
        if isinstance(parameter, ps.IntegerParameter):
            bounds = (parameter.bounds[0], parameter.bounds[1] + 1)
            points_along_dimensions.append([parameter.val2num(v) for v in range(*bounds)])
        elif isinstance(parameter, ps.OrdinalParameter) or isinstance(parameter, ps.CategoricalParameter):
            points_along_dimensions.append([parameter.val2num(v) for v in parameter.values])
        else:
            raise ValueError(f"Only discrete parameters are allowed but got {parameter}")
```

tests/optimizers/botorch_base_test.py
ptighin: True. I think that it was just a mistake. Thanks for spotting this.

tests/optimizers/botorch_base_test.py
ptighin: Great stuff -- thanks! Implemented.

tests/optimizers/botorch_base_test.py
ptighin: I provide a suggestion in the [revised version](https://github.com/boschresearch/blackboxopt/pull/86/files/2b325de9a57b60a358d4a7970c2390490c626720..2932ea21b8de502377e0e1ed9b5d379e271a83c8), please let me know your thoughts :)

blackboxopt/optimizers/botorch_base.py
```diff
@@ -86,12 +118,20 @@ def _acquisition_function_optimizer_factory(
...
+        # choices is not specified
+        choices = torch.from_numpy(
+            _get_numerical_points_from_discrete_space(search_space)
+        ).to(torch_dtype)
+    else:
+        # Optimize over the desired number of samples from the discrete search space
+        choices = torch.Tensor(
+            [
+                search_space.to_numerical(search_space.sample())
+                for _ in range(kwargs.pop("num_random_choices", 5_000))
```
LGro, line 132: Do I see correctly that the default of `5_000` will never be used in the current implementation and could thus be removed? Because if we enter this decision branch, either `num_random_choices in kwargs` or if not, `space_has_continuous_parameters` would need to be true, which would have triggered the return in 111 already.

LGro (approved):
Thanks a lot @ptighin :rocket:

PR #87, opened 2023-02-28
Feature/objective wise none replacement

LGro:
If one wants to use infinity as the numerical loss representation for `None` objective values, the direction of the objective matters. This PR introduces backward compatible support for individual none representation values per objective.

An alternative that I decided against due to the lack of generality and transparency to the call context is detecting `inf` and adding a sign according to the objective direction in the loss vector function.

### DIFF
--- a/blackboxopt/utils.py
+++ b/blackboxopt/utils.py
@@ -6,7 +6,7 @@
 import hashlib
 import pickle
 from itertools import compress
-from typing import Dict, Iterable, List, Optional, Sequence
+from typing import Dict, Iterable, List, Optional, Sequence, Union
 
 import numpy as np
 
@@ -17,7 +17,7 @@ from blackboxopt.evaluation import Evaluation
 def get_loss_vector(
     known_objectives: Sequence[Objective],
     reported_objectives: Dict[str, Optional[float]],
-    none_replacement: float = float("NaN"),
+    none_replacement: Union[float, Sequence[float]] = float("NaN"),
 ) -> List[float]:
     """Convert reported objectives into a vector of known objectives.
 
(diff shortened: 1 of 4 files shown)


blackboxopt/utils.py
```diff
@@ -17,7 +17,7 @@
 def get_loss_vector(
     known_objectives: Sequence[Objective],
     reported_objectives: Dict[str, Optional[float]],
-    none_replacement: float = float("NaN"),
+    none_replacement: Union[float, Sequence[float]] = float("NaN"),
```
sfalkner, line 20: Is there a reason to have a `Sequence[float]` over a `Dict[str, float]`? This would bring it closer to the `reported_objectives` argument, but that's just a minor question.

blackboxopt/utils.py
```diff
@@ -27,16 +27,29 @@ def get_loss_vector(
...
+            respectively.
 
     Returns:
         A list of loss values.
     """
+    if isinstance(none_replacement, Sequence) and (
+        len(none_replacement) != len(known_objectives)
+    ):
+        raise ValueError(
+            f"There are only {len(none_replacement)} None replacement values but there "
```
sfalkner, line 41: Nitpicking alert: This error message would also show up if `none_replacement` is longer than `known_objectives`, but the only indicates that the former is shorter :)

sfalkner (approved):
Thanks @LGro. Looks good, except for the one question I posted.

LGro:
I somehow got confused about objective directions when what we anyway return is losses only, so there really is no need to specify objective direction dependent None replacement values. Closing accordingly, apologies for the confusion @sfalkner 😇

PR #88, opened 2023-03-06
add support for multiple objectives in botorch to numerical

LGro:
Supporting not only one but also multiple objectives in our BOTorch utils function `to_numerical`.

### DIFF
--- a/tests/optimizers/botorch_utils_test.py
+++ b/tests/optimizers/botorch_utils_test.py
@@ -4,7 +4,6 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import numpy as np
-import parameterspace as ps
 import pytest
 import torch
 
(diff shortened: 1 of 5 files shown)


tests/optimizers/botorch_utils_test.py
```diff
@@ -4,6 +4,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import numpy as np
+import parameterspace as ps
```
aeivazi, line 7: are you sure this import is needed? there is no new code that uses `ps`

aeivazi (approved):
Looks great, thank you @LGro!

tests/optimizers/botorch_utils_test.py
LGro: Thanks, will remove :relaxed:

PR #89, opened 2023-03-10
update reference tests to use two objectives in mo case by default

LGro:


LGro:
We're all done now with checking potential implications downstream @aeivazi :relaxed:

aeivazi (approved):
great, thank you!

PR #90, opened 2023-03-24
Feature/replace pylint with ruff

LGro:
This PR replaces pylint with the much faster ruff. Please let me know if you consider the specified rule set to be sufficient or if there are other concerns, @dynobo.
See the separate commits one by one for a smoother review experience.


blackboxopt/visualizations/utils.py
```diff
@@ -13,7 +13,7 @@
 from blackboxopt import Evaluation, Objective
 
 # For backwards compatibility; TODO: remove with next major release
-from blackboxopt.utils import mask_pareto_efficient
+from blackboxopt.utils import mask_pareto_efficient  # noqa
```
dynobo, line 16: Personally, I prefer to ignore only specific rules, e.g. `# noqa: ANN201` or something. Some benefits I see:
- Not (accidentally) masking other errors
- Helps human reader to understand the ignore without the need to run `ruff` (e.g. in PR reviews)
- It helps `ruff` (e.g. if a rule got deprecated, ruff will warn/auto-remove the `noqa`'s for that specific rule)

pyproject.toml
```diff
@@ -72,35 +71,35 @@ profile = "black"
...
-    "too-many-instance-attributes",
-    "too-many-arguments",
-    "too-few-public-methods",
-    "too-many-function-args"
+[tool.ruff]
+# Enable pycodestyle (`E`) and Pyflakes (`F`) codes by default.
+select = ["E", "F"]
+ignore = []
+# Allow autofix for all enabled rules (when `--fix`) is provided.
+fixable = ["A", "B", "C", "D", "E", "F"]
```
dynobo, line 79: How do you feel about adding `"I"` and drop `isort` accordingly?

It could also be useful to add `"RUF"` for applying [ruff specific rules](https://beta.ruff.rs/docs/rules/#ruff-specific-rules-ruf) (especially the `unused noqa` rule is quite useful).

dynobo (approved):
Thanks, @LGro! Looking forward to faster linting... :-)

pyproject.toml
LGro: I'm happy to try it out :relaxed:

LGro: Seems to work, although not 100% consistently with isort, which I don't think is an issue, though. I just noticed for the noqa annotated imports that isort converted into individual imports, ruff combined into one import again.
One downgrade in the current version as of 5c64453 is that I haven't figured out how to tell the ruff isort rules that for `examples/*` the line length is different. However, since this is not an issue right now, I'd propose to solve it when we need to. Unless you already know of a way?

LGro: I'll move forward for now but would be happy to refine the setup as we go :relaxed:

PR #91, opened 2023-03-28
Fix constraints in reference Test

sfalkner:
This PR adds valid constraint values to the first evaluation in the reference test `opimize_single_parameter_sequentially_for_n_max_evaluations`. The behavior for the previous value `None` was not defined.

### DIFF
--- a/blackboxopt/optimizers/testing.py
+++ b/blackboxopt/optimizers/testing.py
@@ -84,11 +84,12 @@ def optimize_single_parameter_sequentially_for_n_max_evaluations(
 
     if issubclass(optimizer_class, MultiObjectiveOptimizer):
         evaluation = eval_spec.create_evaluation(
-            objectives={"loss": None, "score": None}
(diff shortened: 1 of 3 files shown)


blackboxopt/optimizers/testing.py
```diff
@@ -84,11 +84,12 @@ def quadratic_function(p1):
 
     if issubclass(optimizer_class, MultiObjectiveOptimizer):
         evaluation = eval_spec.create_evaluation(
-            objectives={"loss": None, "score": None}
+            objectives={"loss": None, "score": None},
+            constraints={"constraint": 10.0},
```
aeivazi, line 88: but do we have multiobjective **and** constrained optimizer?

aeivazi (approved):

blackboxopt/optimizers/testing.py
LGro: In principle that's a possibility that we should be ready for :relaxed:

PR #93, opened 2023-05-05
update optimizer list in docs

LGro:
I noticed our overview statement about the contained optimizers was a bit outdated, so here is an attempt to update it.


sfalkner (approved):
Thanks @LGro for spotting and fixing it :)

PR #95, opened 2023-05-11
add utils to save and load a study json

LGro:
Before, one would need to take care of storing evaluations with the search space they came from and the matching objectives individually. Now, there are two utility functions for saving and loading to and from JSON respectively.

There is also a corresponding follow up at `feature/save-and-load-study-pickle` in case we like this pattern for JSON.

### DIFF
--- a/tests/utils_test.py
+++ b/tests/utils_test.py
@@ -14,8 +14,10 @@ from blackboxopt.utils import (
     filter_pareto_efficient,
     get_loss_vector,
     load_study_from_pickle,
+    load_study_from_json,
     mask_pareto_efficient,
     save_study_as_pickle,
+    save_study_as_json,
     sort_evaluations,
 )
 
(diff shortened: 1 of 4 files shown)


tests/utils_test.py
```diff
@@ -177,3 +180,26 @@ def test_sort_evaluations_with_different_parameter_order():
...
+
+def test_save_and_load_study_json(tmp_path):
+    tmp_file = tmp_path / "out.json"
+
+    search_space = ps.ParameterSpace()
+    objectives = [Objective("loss", False), Objective("score", True)]
+    evaluations = [
+        Evaluation(configuration={"p1": 1.0}, objectives={"loss": 1.0, "score": 3.0}),
+        Evaluation(configuration={"p1": 2.0}, objectives={"loss": 0.1, "score": 2.0}),
+        Evaluation(configuration={"p1": 3.0}, objectives={"loss": 0.0, "score": 1.0}),
```
dynobo, line 193: Might it be useful, to add a `user_info` with a nested `dict` as value to the test cases? 

(In the past, I experienced that serializing _nested_ objects sometimes cause issues. Probably not an issue for our `Evaluation` dataclass, but I'd test it anyway. Also it might be useful to extend [its description](https://github.com/boschresearch/blackboxopt/blob/main/blackboxopt/evaluation.py#L171) to `...json-serializable...`.)

dynobo (approved):
Cool feature, thanks! :-)

tests/utils_test.py
LGro: Csers can put values with complex types that aren't json serializable into `user_info`, which I'd consider fine and thus wouldn't change the convention for `user_info` to be JSON serializable. I understand this utility function more in the spirit of if it's all simple types, this function can help to and from JSON; and we can document that expected behaviour in an additional test case (e.g. `expect exception when complex type is contained`). There is the alternative of pickling #96 for exactly the cases where complex types are involved.

PR #96, opened 2023-05-11
add utils to save and load a study pickle

LGro:
Same as #95 but for `pickle`.

### DIFF
--- a/blackboxopt/utils.py
+++ b/blackboxopt/utils.py
@@ -4,11 +4,14 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import hashlib
+import os
 import pickle
 from itertools import compress
-from typing import Dict, Iterable, List, Optional, Sequence
+from pathlib import Path
+from typing import Dict, Iterable, List, Optional, Sequence, Tuple
 
 import numpy as np
+import parameterspace as ps
 
 from blackboxopt.base import Objective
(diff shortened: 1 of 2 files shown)


blackboxopt/utils.py
```diff
@@ -99,3 +102,35 @@ def sort_evaluations(evaluations: Iterable[Evaluation]) -> Iterable[Evaluation]:
...
+    objectives: List[Objective],
+    evaluations: List[Evaluation],
+    pickle_file_path: os.PathLike,
+    overwrite: bool = False,
+):
+    """Save space, objectives and evaluations as pickle at `pickle_file_path`."""
+    if Path(pickle_file_path).exists() and not overwrite:
+        raise ValueError(f"{pickle_file_path} exists and overwrite is False")
+
+    with open(pickle_file_path, "wb") as fh:
```
sfalkner, line 118: Not a crucial point, but I believe this to fail if the directory of `pickle_file_path` does not exist. Should it be created in that case or fail with a meaningful message?

sfalkner (approved):
Thanks @LGro, looks very good!

blackboxopt/utils.py
LGro: I went with raising a helpful error and leaving the creation to the user. Thanks!

PR #100, opened 2023-05-22
update torch gpu to cpu project deps (closes #94)

### LINKED ISSUE #94: Revisit ADR on Torch CPU/GPU after Poetry 1.5
With the possibility to define exclusive indexes for poetry starting from version 1.5.0 (python-poetry/poetry/issues/6713 & python-poetry/poetry/issues/7853), let's revisit our [ADR](https://github.com/boschresearch/blackboxopt/blob/main/architecture_decision_records/torch-gpu-vs-cpu-dependency.md) on the matter and potentially switch to cpu torch by default.

LGro:
With the release of poetry 1.5.0 and the according availability of the source attribute `priority = "explicit"` this PR changes the project dependency from the general GPU enabled torch that brings in all kinds of cuda/cudnn dependencies to the CPU version from the respective index. The former downside of poetry update searching the torch index for all dependencies and thus being slow, noted in the torch GPU vs CPU ADR, is resolved by marking the torch index `explicit` and using the `source="pytorch-cpu"` reference for the torch dependency. This should still allow downstream users to install the GPU version of torch instead of the CPU one without issues.


LGro:
The issue with poetry not picking up all pytorch dependencies as seen in our CI/CD here (e.g. missing `sympy`) is related to a torch issue that you are aware of @aeivazi, right? Would you propose limiting torch to `<2.0` for now? And could you link the corresponding torch issue here, please?

LGro:
> The issue with poetry not picking up all pytorch dependencies as seen in our CI/CD here (e.g. missing `sympy`) is related to a torch issue that you are aware of @aeivazi, right? Would you propose limiting torch to `<2.0` for now? And could you link the corresponding torch issue here, please?

I restricted pytorch to <2.0 for now until the corresponding issue is fixed.

aeivazi:
> The issue with poetry not picking up all pytorch dependencies as seen in our CI/CD here (e.g. missing `sympy`) is related to a torch issue that you are aware of @aeivazi, right? Would you propose limiting torch to `<2.0` for now? And could you link the corresponding torch issue here, please?

Reading your explanation, this is not related to `torch==2.0.1`. The issue I was facing was due to installing GPU torch version and missing `nvidia-*` dependencies: https://github.com/pytorch/pytorch/issues/100974.

This error might be specific to CPU torch?...
I've googled the error, there is a recommendation to install `sympy` have you tried? https://github.com/pytorch/vision/issues/7034

LGro:
> This error might be specific to CPU torch?...
> I've googled the error, there is a recommendation to install `sympy` have you tried? [pytorch/vision#7034](https://github.com/pytorch/vision/issues/7034)

True, adding sympy works; I removed the upper bound again. All good from my side then, awaiting your review.

aeivazi (approved):
Awesome addition, thanks @LGro!

PR #103, opened 2023-05-25
remove upper version bound for parameterspace dependency

LGro:
I'd like to propose being less conservative with the version bound of our own package parameterspace to avoid having to adjust the upper bound regularly.


aeivazi (approved):
sounds good!

PR #105, opened 2023-05-31
add on release published as ci/cd trigger

LGro:
It seems like creating releases (and with that a tag) via the GitHub web ui doesn't trigger our CI/CD pipeline anymore. This PR should fix that.


dynobo (approved):
Looks good, thanks!

PR #106, opened 2023-06-01
add multi objective example and default logger init

LGro:
Our only example in the docs so far was the dask distributed one, which is oddly specific and it was a bit lonely. I added a HPO example that shows multiple objectives and a mixed space.


tests/examples_test.py
```diff
@@ -0,0 +1,30 @@
+from pathlib import Path
+
+import pytest
+
+from blackboxopt.examples import multi_objective_multi_param
+
+
+@pytest.mark.parametrize("example_module", [multi_objective_multi_param])
```
lreeb, line 8: Good idea to test the examples! Would you add the other example (dash distributed) in a separate ticket/PR, or would it be easier to include that here in one go?

blackboxopt/utils.py.orig
```diff
@@ -0,0 +1,218 @@
+# Copyright (c) 2020 - for information on the respective copyright owner
```
lreeb, line 1: This file can probably be removed/ignored...

lreeb (approved):
Thanks @LGro!

tests/examples_test.py
LGro: Since it requires a slight rewrite and the PR was already larger than I like, I wanted to handle it separately.

PR #108, opened 2023-06-05
update fixed seed reference test to use two objectives in mo case

LGro:
@aeivazi, I still had this PR in the pipeline from back in March but am not 100% confident if the original motivation still exists - can you help fill in the gaps, whether this is still something we'd like to tackle. I remember we wanted to not expect all multi objective optimizers to successfully handle one objective and thus switch to two objectives in our reference test.

### DIFF
--- a/blackboxopt/optimizers/testing.py
+++ b/blackboxopt/optimizers/testing.py
@@ -150,15 +150,20 @@ def is_deterministic_with_fixed_seed_and_larger_space(
             optimizer_class,
             optimizer_kwargs,
             objective=Objective("loss", False),
-            objectives=[Objective("loss", False)],
+            objectives=[Objective("loss", False), Objective("score", True)],
             seed=seed or 42,
         )
 
         for i in range(n_evaluations):
             es = opt.generate_evaluation_specification()
+
+            objectives = {"loss": losses[i]}
+            if isinstance(opt, MultiObjectiveOptimizer):
+                objectives["score"] = -1.0 * losses[i] ** 2
             evaluation = es.create_evaluation(
-                objectives={"loss": losses[i]}, constraints={"constraint": 10.0}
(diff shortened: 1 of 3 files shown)


blackboxopt/optimizers/testing.py
```diff
@@ -150,15 +150,20 @@ def is_deterministic_with_fixed_seed_and_larger_space(
...
-            objectives=[Objective("loss", False)],
+            objectives=[Objective("loss", False), Objective("score", True)],
             seed=seed or 42,
         )
 
         for i in range(n_evaluations):
             es = opt.generate_evaluation_specification()
+
+            objectives = {"loss": losses[i]}
+            if isinstance(opt, MultiObjectiveOptimizer):
```
aeivazi, line 161: I believe the motivation was to be sure, that MO optimizers could handle more than one objective, was not it? And on my point of view this extension makes a lot of sense. Why are you uncertain?

LGro: Mainly because it was already a couple weeks ago, and I wasn't sure if we landed on some other solution in the meantime, rendering this irrelevant.

LGro: Specifically, I see that we have `optimizer_kwargs` in `main`, which reads to me like this is another solution attempt:
https://github.com/boschresearch/blackboxopt/blob/main/blackboxopt/optimizers/testing.py#L133-L135

aeivazi (approved):
Thanks @LGro

blackboxopt/optimizers/testing.py
aeivazi: Then i believe we've covered MOOptimizers in testing already. Let's then forget about this PR.

LGro:
Closing since the issue is already fixed.

PR #109, opened 2023-06-05
add pre eval callback and stringify user info for logging

### LINKED ISSUE #97: JSON dump in optimization loop logging fails for some user_info values

LGro:
* Close #97 by casting the `user_info` to string before dumping it to JSON for logging
* Add pre-evaluation callback
* Add reference to evaluation function wrapper in docstring for more transparency about `catch_exceptions_from_evaluation_function` argument

### DIFF
--- a/blackboxopt/optimization_loops/sequential.py
+++ b/blackboxopt/optimization_loops/sequential.py
@@ -26,10 +26,11 @@ def run_optimization_loop(
     optimizer: Union[SingleObjectiveOptimizer, MultiObjectiveOptimizer],
     evaluation_function: Callable[[EvaluationSpecification], Evaluation],
     timeout_s: float = float("inf"),
-    max_evaluations: int = None,
+    max_evaluations: Optional[int] = None,
     catch_exceptions_from_evaluation_function: bool = False,
+    pre_evaluation_callback: Optional[Callable[[EvaluationSpecification], Any]] = None,
     post_evaluation_callback: Optional[Callable[[Evaluation], Any]] = None,
-    logger: logging.Logger = None,
+    logger: Optional[logging.Logger] = None,
 ) -> List[Evaluation]:
     """Convenience wrapper for an optimization loop that sequentially fetches evaluation
     specifications until a given timeout or maximum number of evaluations is reached.
@@ -53,9 +54,12 @@ def run_optimization_loop(
             exception raised by the evaluation function or instead store their stack
             trace in the evaluation's `stacktrace` attribute. Set to True if there are
             spurious errors due to e.g. numerical instability that should not halt the
-            optimization loop.
+            optimization loop. For more details, see the wrapper that is used internally
+            `blackboxopt.optimization_loops.utils.evaluation_function_wrapper`
+        pre_evaluation_callback: Reference to a callable that is invoked before each
+            evaluation and takes a `blackboxopt.EvaluationSpecification` as an argument.
         post_evaluation_callback: Reference to a callable that is invoked after each
-            evaluation and takes a `blackboxopt.Evaluation` as its argument.
+            evaluation and takes a `blackboxopt.Evaluation` as an argument.
         logger: The logger to use for logging progress. Default: `blackboxopt.logger`
 
     Returns:
@@ -82,10 +86,13 @@ def run_optimization_loop(
 
         try:
             evaluation_specification = optimizer.generate_evaluation_specification()
+
             logger.info(
-                "The optimizer proposed a specification for evaluation:\n"
-                + f"{json.dumps(evaluation_specification.to_dict(), indent=2)}"
+                "The optimizer proposed the following evaluation specification:\n%s",
+                json.dumps(evaluation_specification.to_dict(), indent=2),
             )
+            if pre_evaluation_callback is not None:
+                pre_evaluation_callback(evaluation_specification)
 
             evaluation = evaluation_function_wrapper(
                 evaluation_function=evaluation_function,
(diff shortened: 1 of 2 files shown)


blackboxopt/optimization_loops/sequential.py
```diff
@@ -82,10 +86,13 @@ def run_optimization_loop(
...
         try:
             evaluation_specification = optimizer.generate_evaluation_specification()
+
             logger.info(
-                "The optimizer proposed a specification for evaluation:\n"
-                + f"{json.dumps(evaluation_specification.to_dict(), indent=2)}"
+                "The optimizer proposed the following evaluation specification:\n%s",
+                json.dumps(evaluation_specification.to_dict(), indent=2),
             )
+            if pre_evaluation_callback is not None:
```
dynobo, line 94:
```suggestion
            if pre_evaluation_callback:
```

blackboxopt/optimization_loops/sequential.py
```diff
@@ -94,16 +101,25 @@ def run_optimization_loop(
                 objectives=objectives,
                 catch_exceptions_from_evaluation_function=catch_exceptions_from_evaluation_function,
             )
+
             logger.info(
-                "Reporting the result from the evaluation function to the optimizer:\n"
-                + f"{json.dumps(evaluation.to_dict(), indent=2)}"
+                "Reporting the following evaluation result to the optimizer:\n%s",
+                json.dumps(
```
dynobo, line 107: As we here use it only for logging, might stringifying via [`pprint.pformat`](https://docs.python.org/3/library/pprint.html#pprint.pformat) be an option? Might hope would be, that it already handles most issues that can occur when serializing to json...

blackboxopt/optimization_loops/sequential.py
```diff
@@ -53,9 +54,12 @@ def run_optimization_loop(
             exception raised by the evaluation function or instead store their stack
             trace in the evaluation's `stacktrace` attribute. Set to True if there are
             spurious errors due to e.g. numerical instability that should not halt the
-            optimization loop.
+            optimization loop. For more details, see the wrapper that is used internally
+            `blackboxopt.optimization_loops.utils.evaluation_function_wrapper`
+        pre_evaluation_callback: Reference to a callable that is invoked before each
```
dynobo, line 59: I'm wondering: What's actually the use case for this pre-evaluation callback? And if I'm asking myself this question, might it make sense to add such an example to the docstring, too?

dynobo (approved):
Looks fine, I'm only curious, for which use-cases this feature is used. :-)

sfalkner (approved):
Thanks @LGro. These changes make the optimization loop more flexible 👍

blackboxopt/optimization_loops/sequential.py
LGro: With `str()` and `json.dumps` as currently implemented:
```
  "created_unixtime": 1686052139.35081,
  "context": null,
  "constraints": null,
  "finished_unixtime": 1686052139.351386,
  "stacktrace": null,
  "user_info": "{'muy': 'complex', 'lamda': <function evaluation_function.<locals>.<lambda> at 0x7f08e22c9790>, 'longlist': [2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2]}"
}
```

With `pprint.pformat(evaluation.to_dict())` instead of `json.dumps`
```
 'created_unixtime': 1686052409.133704,
  'finished_unixtime': 1686052409.134927,
  'objectives': {'loss': 1.0},
  'optimizer_info': {},
  'settings': {},
  'stacktrace': None,
  'user_info': { 'lamda': <function evaluation_function.<locals>.<lambda> at 0x7f0964e23b80>,
                 'longlist': [ 2,
                               2,
                               2,
                               2,
                               2,
                               2,
                               2,
                               2,
                               2,
                               2,
                               2,
                               2,
                               2,
                               2,
                               2,
                               2,
                               2,
                               2,
                               2,
                               2],
                 'muy': 'complex'}}
```

with `pprint.pformat()` just on the `user_info` and around `json.dumps`:
```
  "created_unixtime": 1686052342.399883,
  "context": null,
  "constraints": null,
  "finished_unixtime": 1686052342.400269,
  "stacktrace": null,
  "user_info": "{'lamda': <function evaluation_function.<locals>.<lambda> at 0x7f64ad4aa040>,\n 'longlist': [2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2],\n 'muy': 'complex'}"
}
```

To me the current implementation looks the cleanest, or am I missing specific pprint args?

dynobo: > or am I missing specific pprint args? 

well, `compact=True` should help, could you try that?

Probably it still wouldn't look exactly like the custom solution, on the other hand it also wraps lines after _n_ chars etc... But I see the simpler implementation as main benefit.

LGro: With compact it does look a wee bit nicer
```
{'configuration': {'p': 0.40032391669228673},
 'constraints': None,
 'context': None,
 'created_unixtime': 1686055220.515068,
 'finished_unixtime': 1686055220.516517,
 'objectives': {'loss': 1.0},
 'optimizer_info': {},
 'settings': {},
 'stacktrace': None,
 'user_info': {'lamda': <function evaluation_function.<locals>.<lambda> at 0x7f9d4fda9820>,
               'longlist': [2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2,
                            2, 2, 2],
               'muy': 'complex'}}
```

What do the others think, @aeivazi and @sfalkner? I'm open to changing to pprint, or has someone of you benefited from being able to copy valid JSON from the log output?

blackboxopt/optimization_loops/sequential.py
dynobo: @LGro, makes sense, thanks!

blackboxopt/optimization_loops/sequential.py
dynobo: > or has someone of you benefited from being able to copy valid JSON from the log output?

What difference do you see regarding that with `pformat()` vs. `json.dumps()`? By allowing non-serializable objects, both never will be valid JSONs...

LGro: The difference is that the initially proposed `json.dumps` with `str(user_info)` results in valid JSON, it just changes the `user_info` value into a string, while pformat does not. But at least I haven't relied on that so far, just wanted to check because it's the only reason left for me why I'd consider json over pprint.

aeivazi: Honestly, i have not yet used log output for anything else but additional information. I feel both proposal yields to the same result for me as a user. If `pformat` reduces number of code lines, i would vote for it.

PR #110, opened 2023-06-05
add test for dask distributed example

LGro:
* Wrap the dask distributed example in a `main`
* Add dask example to tests
* Remove superfluous integration test marker


lreeb (commented):
Why do the tests fail? Is it a problem with running them on the github actions runner?

LGro:
> Why do the tests fail? Is it a problem with running them on the github actions runner?

That's unclear to me so far, I'm suspecting an issue with dask opening up ports on localhost and trying to connect locally via tcp/ip. If you have time, feel free to take a look as well.

PR #112, opened 2023-06-05
add pipefail for tests in action (closes #111)

### LINKED ISSUE #111: Passing Actions workflow despite failing tests
With our current
```
 poetry run pytest --cov=blackboxopt tests | sed -n '/ -----------/,$p' > coverage.txt
```
failing tests don't cause the GH Actions workflow step to fail anymore.

LGro:
Seems like failing tests didn't cause our GH Actions workflow to fail anymore ever since I added the pipe for the coverage report (see two commits in this PR).

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -48,7 +48,10 @@ jobs:
           poetry install --extras "all"
       - name: Run tests
         if: always()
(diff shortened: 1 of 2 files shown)


LGro:
Seems just like the default shell `sh` doesn't support the `pipefail` option. I'll see if we can switch to bash or find an alternative.

PR #113, opened 2023-06-05
replace custom coverage action with py-cov-comment

LGro:
Our self made coverage reporting wasn't only causing failing tests to slip (#111) but also hid the errors when they occurred. So here is an attempt to switch to https://github.com/marketplace/actions/python-coverage-comment as recommended by @dynobo.


dynobo (approved):
:rocket:

PR #114, opened 2023-06-07
expose num evaluations in reference test, fix docstrings

LGro:
Introduce `n_max_evaluations` argument for `handles_conditional_space` like we already had for `optimize_single_parameter_sequentially_for_n_max_evaluations` in case users only want to try fewer evaluations. This can be helpful if the optimizer is slow and one can be confident that fewer evaluations are sufficient proof. The default behaviour remains unchanged.

Also, while I noticed, I removed the outdated return statements in the reference tests docstrings.


lreeb (approved):
Thanks @LGro !

PR #116, opened 2023-07-03
remove redundant non dominated sorting in hypervolume vis

LGro:
Turns out, the pymoo hypervolume indicator by default does ensure non dominated sorting based filtering of the input loss matrix to only contain the pareto front. Accordingly, our manual additional pareto front filtering is gone now and I also made the default `nds=True` explicit in the indicator initialization. From anecdotal evidence, this yields in a speed up of ~10x.

We can likely speed the sequential hypervolume computation up even further with a custom implementation based on our pareto front util function along with switching to `nds=False`, but that is for a separate PR.

### DIFF
--- a/blackboxopt/__init__.py
+++ b/blackboxopt/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "5.0.0"
+__version__ = "5.0.1"
 
 from parameterspace import ParameterSpace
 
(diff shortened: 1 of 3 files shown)


aeivazi (approved):
Thank you @LGro!

PR #117, opened 2023-07-19
refactor botorch base predict_model_based_best into function

LGro:
Moving the `predict_model_based_best` method logic into a pure function, leaving the method as a user facing wrapper for convenience. This is in preparation for the upcoming botorch base refactoring to enable multi objective optimizers more easily.

### DIFF
--- a/blackboxopt/optimizers/botorch_utils.py
+++ b/blackboxopt/optimizers/botorch_utils.py
@@ -6,8 +6,10 @@
 import warnings
 from typing import Iterable, List, Optional, Sequence, Tuple
 
+import botorch.models.model
 import numpy as np
 import parameterspace as ps
+import scipy.optimize as sci_opt
 import torch
 from sklearn.impute import SimpleImputer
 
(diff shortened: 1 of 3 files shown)


sfalkner (approved):
Thanks @LGro. Hopefully, we get to refactor some more soon :)

dynobo (approved):
Looks good! :+1:

blackboxopt/optimizers/botorch_utils.py
```diff
@@ -232,7 +232,8 @@ def posterior_mean(x):
             jac=False,
             x0=x,
             args=(),
-            bounds=bounds,
+            # The numerical representation always lives on the unit hypercube
+            bounds=torch.Tensor([[0, 1]] * len(search_space)).to(dtype=torch_dtype),
```
sfalkner, line 236: According to the [pytorch docs](https://pytorch.org/docs/stable/tensors.html#torch.Tensor), we could also use
`bounds=torch.tensor([[0, 1]] *len(search_space), dtype=torch_dtype)`, which is the listed way of creating a tensor with pre-existing data.

sfalkner (approved):
Looks good from my side. Thanks @LGro !

blackboxopt/optimizers/botorch_utils.py
LGro: We've seen pylint struggle with the `torch.tensor()` interface, claiming "Module 'torch' has no 'tensor' member" but I haven't figured out why. To keep the noise down in case someone is using pylint extensions in their IDE, I'd stick with the current proposal, assuming that the overhead is negligible here.

PR #118, opened 2023-08-03
Hotfix/random proposals until valid evals

LGro:
If `num_initial_random_samples` is specified > 0, keep sampling random specifications until `num_initial_random_samples`  many *valid* evaluations were reported.

### DIFF
--- a/blackboxopt/optimizers/botorch_base.py
+++ b/blackboxopt/optimizers/botorch_base.py
@@ -283,9 +283,11 @@ class SingleObjectiveBOTorchOptimizer(SingleObjectiveOptimizer):
             # at least one of them has a valid objective
             eval_spec = EvaluationSpecification(
                 configuration=self.search_space.sample(),
+                optimizer_info={"model_based_pick": False},
             )
         else:
             eval_spec = self._generate_evaluation_specification()
+            eval_spec.optimizer_info["model_based_pick"] = True
 
         eval_id = self.X.size(-2) + len(self.pending_specifications)
         eval_spec.optimizer_info["evaluation_id"] = eval_id
--- a/blackboxopt/optimizers/botorch_base.py
+++ b/blackboxopt/optimizers/botorch_base.py
@@ -278,15 +278,11 @@ class SingleObjectiveBOTorchOptimizer(SingleObjectiveOptimizer):
         ):
             raise OptimizerNotReady
 
+        # Generate random samples until there are enough samples where at least one of
+        # the objective values is available
         if self.num_initial_random > 0 and (
-            self.X.size(-2) < self.num_initial_random
-            or (
-                torch.nonzero(~torch.any(self.losses.isnan(), dim=1)).numel()
-                < self.num_initial_random
-            )
+            sum(~torch.any(self.losses.isnan(), dim=1)) < self.num_initial_random
         ):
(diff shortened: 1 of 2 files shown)


tests/optimizers/botorch_base_test.py
```diff
@@ -166,6 +166,56 @@ def test_find_optimum_in_1d_discrete_space(seed):
...
+        es.create_evaluation(objectives={"loss": None}),
+    )
+
+    es = opt.generate_evaluation_specification()
+    assert not es.optimizer_info[
+        "model_based_pick"
+    ], "One valid evaluation reported, 1 < 2 initial random samples"
+    opt.report(
+        es.create_evaluation(objectives={"loss": es.configuration["integ"] ** 2}),
+    )
```
LGro, line 211: Before the PR, this would have already been a model based pick.

tests/optimizers/botorch_base_test.py
```diff
@@ -166,6 +166,56 @@ def test_find_optimum_in_1d_discrete_space(seed):
...
+        "model_based_pick"
+    ], "One evaluation reported, 1 < 2 initial random samples"
+    opt.report(
+        es.create_evaluation(objectives={"loss": None}),
+    )
+
+    es = opt.generate_evaluation_specification()
+    assert not es.optimizer_info[
+        "model_based_pick"
+    ], "One valid evaluation reported, 1 < 2 initial random samples"
```
LGro, line 208: Before this PR, this would have already been a model based pick.

blackboxopt/optimizers/botorch_base.py
```diff
@@ -277,15 +280,20 @@ def generate_evaluation_specification(self) -> EvaluationSpecification:
 
         if self.num_initial_random > 0 and (
             self.X.size(-2) < self.num_initial_random
-            or torch.nonzero(~torch.any(self.losses.isnan(), dim=1)).numel() == 0
+            or (
+                torch.nonzero(~torch.any(self.losses.isnan(), dim=1)).numel()
+                < self.num_initial_random
+            )
```
sfalkner, line 286: Note, that the comment below now is wrong, because sampling random configurations now continues until `num_initial_random` many have all valid objectives.

blackboxopt/optimizers/botorch_base.py
```diff
@@ -277,15 +280,20 @@ def generate_evaluation_specification(self) -> EvaluationSpecification:
 
         if self.num_initial_random > 0 and (
             self.X.size(-2) < self.num_initial_random
```
sfalkner, line 282: I believe this line is redundant. Both `self.losses` and `self.X` are always of the same size and this line is always satisfied, if the condition below is satisfied. Or am I missing a corner case here?

tests/optimizers/botorch_base_test.py
sfalkner: True, but it was consistent with the behavior described in the in-line comment.
But I like this solution now much better.

sfalkner (approved):
Thanks @LGro for making this more consistent.

blackboxopt/optimizers/botorch_base.py
LGro: You're right, I'll simplify accordingly.

PR #119, opened 2023-08-10
relax pymoo bound to 0.5 using backward compat hypervolume api

LGro:
While it's nice that we can just re-use pymoos hypervolume implementation, this introduces a rather specific optimization related dependency for our visualization extra. To allow for compatibility with a wider range of pymoo versions, courtesy of pymoo's diligence regarding backward compatibility, this PR switches to an older (hypervolume) indicator api that is still supported and thus allows compatibility with pymoo 0.5 as well as the most recent 0.6.

### DIFF
--- a/blackboxopt/__init__.py
+++ b/blackboxopt/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "5.0.1"
+__version__ = "5.0.2"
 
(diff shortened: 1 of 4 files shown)


sfalkner (approved):
Thanks @LGro for taking care of this.

PR #120, opened 2023-10-05
add py3.12 support

LGro:
Adding support for Python 3.12. The main concern might be that we need a very recent pytorch version for that, which might cause incompatibilities with other third party dependencies. I'd suggest we try it out :relaxed:


.github/workflows/python.yaml
```diff
@@ -90,8 +90,8 @@ jobs:
   deploy:
     needs: [lint, test]
     if: startsWith(github.ref, 'refs/tags/v')
-    runs-on: ubuntu-20.04
-    container: python:3.8-slim
+    runs-on: ubuntu-24.04
+    container: python:3.11
```
aeivazi, line 97: no slimness is intentional?

aeivazi (approved):
thank you, just one comment on PR ;)

.github/workflows/python.yaml
LGro: Yes, we need gcc (also noted in the commit message for https://github.com/boschresearch/blackboxopt/pull/120/commits/9cf66de3a33778eff3e6f47a81187fe28c4af4f6, which I'll include in the squashed commit).

PR #121, opened 2023-10-31
Run reporting step also on main

### LINKED ISSUE #115: Enable coverage comparison
With the coverage reporting action we still get the following note:
> No coverage data of the default branch was found for comparison. A possible reason for this is that the coverage action has not yet run after a push event and the data is therefore not yet initialized.

Enable storing the coverage in the respective branch to allow the comparison in each PR.

Could it be that the following line points to a non existing step output because the step names changed between the example and our implementation?
https://github.com/boschresearch/blackboxopt/blob/main/.github/workflows/python.yaml#L85

dynobo:
_Should_ fix #115, but it can only be verified once merged to `main`.

### DIFF
--- a/.github/workflows/python.yaml
+++ b/.github/workflows/python.yaml
@@ -3,10 +3,10 @@ name: ci-cd-pipeline
 on:
   push:
     branches:
-    - main
+      - main
   pull_request:
     branches:
-    - main
+      - main
   release:
     types: [published]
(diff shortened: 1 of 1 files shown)


LGro (approved):
Alright, let's try and keep a close eye on it so that we can revert if it behaves unexpectedly :relaxed:

PR #122, opened 2023-11-20
callbacks for dask optimization loop

sfalkner:
- add pre- and post-evaluation callbacks to dask optimization loop
- update mypy due to bug with numpy code


blackboxopt/optimization_loops/dask_distributed.py
```diff
@@ -164,6 +174,11 @@ def run_optimization_loop(
                 break
 
         new_evaluations = dask_scheduler.check_for_results(timeout_s=20)
+
+        if post_evaluation_callback:
+            for e in new_evaluations:
+                post_evaluation_callback(e)
```
dynobo, line 180: What do you think about
```suggestion
        map(post_evaluation_callback, new_evaluations)
```
or
```suggestion
        _ = map(post_evaluation_callback, new_evaluations)
```
?

dynobo (approved):
Thanks, @sfalkner , seems like a very useful feature addition :-)

blackboxopt/optimization_loops/dask_distributed.py
dynobo: I messed up this suggestion. I'll create a new comment.

dynobo, line 180: What do you think about this instead?
```suggestion
            list(map(post_evaluation_callback, new_evaluations))
```

LGro (approved):
For the sequential loop we have pre and post evaluation callback tests and I think this would also be a good idea here to ensure that we don't accidentally break this API and also to demonstrate the expected behaviour that the order of the evaluations is not necessarily consistently ordered.

PR #123, opened 2024-02-06
add fidelities calculator for staged iteration optimizers

LGro:
Whenever I speak with folks using BOHB or Hyperband, they struggle to grasp the connection between the fidelity related settings and how they translate to the actual fidelity levels that are evaluated. This little JavaScript calculator in our docs should help with that, I hope.
I was contemplating whether to go for a table or flexbox with div containers, and concluded that the form is so small, that mobile device optimization is not necessary, wdyt?


sfalkner (approved):
Thanks @LGro for this helpful addition. Let's hope it renders correctly on GH pages :)

PR #124, opened 2024-02-06
add file based distributed optimization loop

LGro:
For a distributed system where a file system is shared, this introduces an optimization loop setup where the optimizer always keeps the batch of proposals filled and clients can read proposed eval specs and report corresponding results.


blackboxopt/optimization_loops/file_based_distributed.py
```diff
@@ -0,0 +1,160 @@
...
+        max_evaluations=max_evaluations, timeout_s=timeout_s, logger=logger
+    )
+
+    start = time.time()
+    while True:
+        evaluations_to_report = glob.glob(str(target_directory / "eval_result_*.json"))
+        for eval_result in evaluations_to_report:
+            with open(eval_result, "r", encoding="utf-8") as fh:
+                evaluation = Evaluation(**json.load(fh))
+            Path(eval_result).unlink(missing_ok=True)
```
sfalkner, line 70: Why is it ok, if the file is missing? Is this simply for robustness, or is there another mechanism that should delete this file?

blackboxopt/optimization_loops/file_based_distributed.py
```diff
@@ -0,0 +1,160 @@
...
+
+    _max_evaluations = init_max_evaluations_with_limit_logging(
+        max_evaluations=max_evaluations, timeout_s=timeout_s, logger=logger
+    )
+
+    start = time.time()
+    num_evaluations = 0
+    while time.time() - start < timeout_s and num_evaluations < _max_evaluations:
+        current_proposals = glob.glob(str(target_directory / "eval_spec_*.json"))
+        for eval_spec_path in current_proposals:
```
sfalkner, line 141: Looping over this list runs into the risk that a worker might try to load a file that existed before the previous evaluation started, but was already processed in the mean time by a different worker. In that case, the `open` will fail in the next line.
Not that there is a 100% safe way of doing it, but I would propose to only process a single proposal from the list (e.g. pick a random one) and update it every time before a file is loaded. If this needs to be more robust, an additional `try` `except` block could catch the (hopefully) rare cases where the file is deleted between the `glob` and the `open`.

blackboxopt/optimization_loops/file_based_distributed.py
```diff
@@ -0,0 +1,160 @@
...
+)
+
+
+def run_optimization_loop(
+    optimizer: Union[SingleObjectiveOptimizer, MultiObjectiveOptimizer],
+    target_directory: PathLike,
+    timeout_s: float = float("inf"),
+    max_evaluations: Optional[int] = None,
+    proposal_queue_size: int = 1,
+    logger: Optional[logging.Logger] = None,
```
sfalkner, line 35: Should this optimization loop also have the callbacks like the [sequential one](https://github.com/boschresearch/blackboxopt/blob/main/blackboxopt/optimization_loops/sequential.py)? For logging or something similar.

PR #126, opened 2024-03-01
update actions to avoid coverage report on release

LGro:
This should fix the actions error we saw in our previous release: https://github.com/boschresearch/blackboxopt/actions/runs/7916729063


sfalkner (approved):
Thanks @LGro for figuring this out. Much appreciated!

PR #134, opened 2025-12-09
add python 3.14 support

LGro:
The trade-off for supporting Python 3.14 seems to be that we need to raise the lower version bound on SciPy, which in turn requires us to drop support for still supported Python 3.10. I've now specified the SciPy version bounds depending on the Python version to cover 3.10 and 3.14


dynobo (approved):
Thanks, @LGro

PR #135, opened 2026-04-23
switch from poetry to uv

14e6:
Using uv speeds up dependency resolution especially with the broad range of Python versions we support and heavy weight dependencies like torch.


pyproject.toml
```diff
@@ -1,11 +1,11 @@
-[tool.poetry]
+[project]
 name = "blackboxopt"
-version = "5.3.4"
+version = "5.3.5"
```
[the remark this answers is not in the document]
14e6: @copilot apply changes based on [this feedback](https://github.com/boschresearch/blackboxopt/pull/135#discussion_r3129652671)

blackboxopt/optimization_loops/dask_distributed.py
```diff
@@ -43,6 +43,7 @@ def __init__(
         self.objectives = objectives
         self.logger = logger
         self._not_done_futures: Set = set()
+        self._futures_eval_specs: dict = {}
```
sfalkner, line 46: Did mypy always complain about that, or was there a change recently?

sfalkner (approved):
Thanks, @14e6 for transitioning the package to `uv` and fixing some minor things along the way :)

blackboxopt/optimization_loops/dask_distributed.py
14e6: This only popped up now. I haven't investigated how different Python versions and a potential mypy update have caused this, but found it a fair complaint and thus went for the separate book keeping. Since we pop the values this should also not risk any unnecessary information storage.

PR #136, opened 2026-04-28
fix botorch base optimizer fantasize observation noise argument change

14e6:


blackboxopt/optimizers/botorch_base.py
```diff
@@ -230,7 +230,7 @@ def _create_fantasy_model(self, model: Model) -> Model:
         model = model.fantasize(  # type: ignore[operator]
             pending_X,
             IIDNormalSampler(torch.Size((1,))),
-            observation_noise=False,
+            observation_noise=None,
```
14e6, line 233: This is the key change in this PR ☺️

tests/optimizers/botorch_base_test.py
```diff
@@ -56,6 +56,57 @@ def test_all_reference_tests(reference_test, seed):
...
+        ),
+        max_pending_evaluations=5,
+    )
+    es1 = opt.generate_evaluation_specification()
+    opt.report(es1.create_evaluation(objectives={"loss": 0.5}))
+    es2 = opt.generate_evaluation_specification()
+    # Trigger fantasize model
+    es3 = opt.generate_evaluation_specification()
+    assert es1.configuration != es2.configuration
+    assert es1.configuration != es3.configuration
```
14e6, line 84: This missing test caused us to miss this issue due to insufficient coverage of the fantasize call.

blackboxopt/optimizers/botorch_base.py
sfalkner: Thanks for pointing this out, that helped a lot actually :)

tests/optimizers/botorch_base_test.py
```diff
@@ -56,6 +56,57 @@ def test_all_reference_tests(reference_test, seed):
...
+
+
+def test_not_ready():
```
sfalkner, line 87: Thanks for also adding this test, even though it is not 100% related to the issue :)

sfalkner (approved):
Thanks @14e6 for fixing this and the additional test :)

PR #137, opened 2026-05-05
fix predict model based best on empty model

14e6:
Due to the recent botorch update we could run into an issue we didn't previously catch when trying to predict best with an empty model, so I added a test and fix for it.


sfalkner (approved):
Thanks @14e6 for spotting and immediately fixing it :)

PR #138, opened 2026-05-06
Add visualization with objective over iterations including constraint violation indicators

14e6:
Adding a new visualization to our catalog, looks something like this:
image

### DIFF
--- a/tests/visualization_test.py
+++ b/tests/visualization_test.py
@@ -545,8 +545,8 @@ def test_get_constraint_satisfaction_mask():
     mask = get_constraint_satisfaction_mask(evaluations)
     np.testing.assert_array_equal(mask, [True, True, True])
 
-    # c1 >= 0
-    mask = get_constraint_satisfaction_mask(evaluations, {"c1": (0.0, None)})
+    # c2 <= 0
+    mask = get_constraint_satisfaction_mask(evaluations, {"c2": (None, 0.0)})
     np.testing.assert_array_equal(mask, [True, False, False])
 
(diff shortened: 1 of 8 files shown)


14e6:
I considered adding multi-fidelity support but that would have meant I need to use the same or similar color / different alpha to indicate the infeasible points in addition to making them x vs o. If you prefer that trade-off the other way around, I'm happy to support multiple fidelities.

tests/visualization_test.py
```diff
@@ -460,3 +462,151 @@ def test_parallel_coordinate_plot_parameters_with_unhashable_parameters():
...
+        ),
+    ]
+
+    # No bounds: all feasible
+    mask = get_constraint_satisfaction_mask(evaluations)
+    np.testing.assert_array_equal(mask, [True, True, True])
+
+    # c1 >= 0
+    mask = get_constraint_satisfaction_mask(evaluations, {"c1": (0.0, None)})
+    np.testing.assert_array_equal(mask, [True, False, False])
```
sfalkner:
```suggestion
    # c2 <= 0
    mask = get_constraint_satisfaction_mask(evaluations, {"c2": (None,0.0)})
    np.testing.assert_array_equal(mask, [True, False, False])
```
May I suggest a constraint with only an upper bound to also cover that case, as it triggers a different `if` statement? Not that I see anything wrong with the implementation, just for completeness :)

sfalkner (approved):
Thanks @14e6 for this nice visualization.

PR #139, opened 2026-05-07
switch from random to sobol sampling for GPBO random initial samples

14e6:


blackboxopt/optimizers/botorch_base.py
```diff
@@ -189,6 +190,10 @@ def __init__(
         """
         super().__init__(search_space=search_space, objective=objective, seed=seed)
         self.num_initial_random = num_initial_random_samples
+        # Initialize sobol sampler as random initial sampler; the objective is not used.
+        self._random_sampler = SpaceFilling(
+            objectives=[objective], search_space=search_space, seed=seed
+        )
```
14e6, line 196: With the two options of either duplicating the SpaceFilling Sobol initialization and ParameterSpace transformation or using the full optimizer with the unnecessary objective argument here, I picked the latter because I judged the risk of a breaking SpaceFilling optimizer signature change to be low enough and value the consistency across the our different places where we do scrambled and correctly seeded sobol sampling.

sfalkner: I fully agree!

sfalkner (approved):
Thanks @14e6 for taking the time to implement this!

